PR #80, opened 2018-04-17
Fix infinite recursive call

moorereason:
From megacheck: infinite recursive call (SA5007)


xcoulon:
thanks again, @moorereason !

PR #81, opened 2018-04-17
types: tidy up initials func

moorereason:
More megacheck findings:

types/grammar_types.go:279:12: the argument is already a string, there's no need to use fmt.Sprintf (S1025)
types/grammar_types.go:280:2: unnecessary nil check around range (S1031)

### DIFF
--- a/types/grammar_types.go
+++ b/types/grammar_types.go
@@ -276,11 +276,9 @@ func NewDocumentAuthor(namePart1, namePart2, namePart3, emailAddress interface{}
 }
 
 func initials(firstPart string, otherParts ...string) string {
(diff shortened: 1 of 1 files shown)


xcoulon:
thanks a lot, @moorereason!

xcoulon:
@moorereason FYI I just amended the commits of your 2 PR to follow the [Angular Commit Message Convention](https://gist.github.com/stephenparish/9941e89d80e2bc58a153#commit-message-conventions), so you may have to `pull --force` if you got the initial commits in your local `master` branch. Sorry for the trouble I may have caused you.

PR #84, opened 2018-04-20
Admonition fix

raghur:
Hi - so I found an issue where block admonition isn't being parsed properly. It's picked up as an elementattribute rather than admonition block. Adding a negation check to the ElementAttribute fixes it but I'm puzzled how the unit test in admonition_test.go is passing :(

Here's the test file I'm using
```
[CAUTION]                      
ddddd sldfjas asfdsf           
                               
                               
[NOTE]                         
.Title2                        
====                           
This is second admonition      
                               
with another para              
====                           
```

I also added an option for setting log level from the cli. Here's a log of the parse with master
https://pastebin.com/3KkF1wjn

and here's one with the fix:
https://pastebin.com/Xn38RN53

PS - I also added parsing an admonition block with a DelimitedBlock - though that's not relevant. I'm wondering how to write a test case for the original issue :)...

### DIFF
--- a/cmd/libasciidoc/root_cmd.go
+++ b/cmd/libasciidoc/root_cmd.go
@@ -6,8 +6,11 @@ import (
 
 	"github.com/bytesparadise/libasciidoc"
 	"github.com/bytesparadise/libasciidoc/renderer"
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
+
+var logLevel string
 
 // NewRootCmd returns the root command
 func NewRootCmd() *cobra.Command {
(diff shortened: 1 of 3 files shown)


xcoulon:
thanks for your help @raghur! May I ask you to make a distinct PR for the log level in the command line  (ideally, using the https://gist.github.com/stephenparish/9941e89d80e2bc58a153#commit-message-conventions for the commit msg), please, so keep things separated (1 smaller commit per task/bug fix) ? I'll give a look at the admonition grammar and try to understand why/how the test passed in the first place. I'd like to have a test that verifies the fix for good, actually.

raghur:
Yes of course... I don't expect this PR to be merged. Still feeling my way around the source and ran into this.

PR #85, opened 2018-04-20
feat(cmd): add support to specify log level

raghur:
adds flag `--log` as follows

```
 --log string      log level to set {debug, info, warning, error, fatal, panic} (default "warning")
```


xcoulon:
thanks for this PR, @raghur!

Issue #83, opened 2018-04-18
Add flag to only generate the HTML body


xcoulon:

xcoulon:
Fixed in https://github.com/bytesparadise/libasciidoc/pull/95

PR #95, opened 2018-04-24
feat(cmd): add flag to suppress header/footer

raghur:
see #83 is this what this issue refers to?


xcoulon:
thanks for this PR, @raghur!

PR #98, opened 2018-05-01
refactor(parser/renderer): refactor admonition paragraph, support admonition blocks

### LINKED ISSUE #93: Support admonition blocks
the following syntax is used for admonition blocks, i.e. when there's more than 1 paragraph in the admonition:

````
[NOTE]
====
...
====
````

where the content can be any valid asciidoc element.

xcoulon:
remove `Admonition` specific type and instead, treat the admonition attributes
as regular attributes on paragraphs and example blocks, but use a specific template
at the rendering stage.

also, convert test names to lowecase

Fixes #93


xcoulon:
@raghur FYI, this PR should fix the problem you reported a while ago with regards to admonition blocks (see test https://github.com/bytesparadise/libasciidoc/pull/98/files#diff-5b6524e61ace7fefe249c81320830212R119). Only difference with Asciidoctor is the way trailing spaces are trimmed or not.

xcoulon:
@raghur ah, it looks like your https://github.com/bytesparadise/libasciidoc/blob/master/cmd/libasciidoc/test/admonition.adoc has some syntax that is not supported yet, so I'll fix that in this PR ;)

PR #154, opened 2018-07-10
Updating builde instructions in README

etsauer:
Updating the main README with references to build commands and dependencies.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -32,7 +32,14 @@ Further elements will be supported in the future. Feel free to open issues https
 
 == Installation
 
+First, install build dependencies:
+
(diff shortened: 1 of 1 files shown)


xcoulon:
thanks for this PR, @etsauer, but I think that we could make it even better:
- firstly, there's an `install` goal in the `makefile` that builds the binary and copies it into `$GOPATH/bin`
- secondly, we could include goals in the `makefile` to install `dep` and install/update `mna/pigeon`, so it's transparent to the user.

What do you think about it?

etsauer:
@xcoulon yes, I agree. I'll work on making those changes.

xcoulon:
solved in #222

PR #171, opened 2018-08-22
feat(parser/renderer): support role attributes, refactor attributes and image type

### LINKED ISSUE #151: Improve support for `id`, `title` and `role` attributes
The `role` attributes is defined using the `[.role]` syntax.
The `id` and `title` attributes should be handled in the images and lists.

xcoulon:
While improving support for role, id and title in the lists, this patch
also introduces changes in the attributes which get their own type
instead of `map[string]interface{}`, which allows for utility method(s).

Also, pursuing on images, the `Macro` type that is common to `BlockImage`
and `InlineImage` is behind removed, to avoid having to deal with 2 levels
of attributes for the same element.

Fixes #151


pkg/renderer/html5/renderer.go
```diff
@@ -225,6 +225,21 @@ func getID(attributes map[string]interface{}) string {
 	return id
 }
 
+// hasRole checks if the given map has an entry with key `types.AttrRole`
+func hasRole(attributes map[string]interface{}) bool {
```
golangci: `hasRole` is unused

pkg/renderer/html5/renderer.go
```diff
@@ -225,6 +225,21 @@ func getID(attributes map[string]interface{}) string {
...
+	_, found := attributes[types.AttrRole]
+	return found
+}
+
+// getRole returns the value for the entry with key `types.AttrRole` in the given map
+func getRole(attributes map[string]interface{}) string {
```
golangci: `getRole` is unused

pkg/types/grammar_types.go
```diff
@@ -1658,71 +1678,81 @@ func NewElementAttributes(attributes []interface{}) map[string]interface{} {
...
+func NewAdmonitionAttribute(k AdmonitionKind) (ElementAttributes, error) {
+	return ElementAttributes{AttrAdmonitionKind: k}, nil
 }
 
 // NewAttributeGroup initializes a group of attributes from the given generic attributes.
-func NewAttributeGroup(attributes []interface{}) (map[string]interface{}, error) {
+func NewAttributeGroup(attributes []interface{}) (ElementAttributes, error) {
 	// log.Debugf("initializing a new AttributeGroup with %v", attributes)
-	result := make(map[string]interface{}, 0)
+	result := make(ElementAttributes, 0)
```
golangci: should use make(ElementAttributes) instead

PR #216, opened 2018-10-16
feat(make): add linters target (#61)

### LINKED ISSUE #61: Include other go tools (`gofmt`, etc.) in the Makefile

moorereason:
Add linters target to run gofmt and golint checks against codebase.

Fixes #61

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,6 @@ COVERPKGS := $(shell go list ./... | grep -v vendor | paste -sd "," -)
 
 DEVTOOLS=\
 				github.com/golang/dep/cmd/dep \
-				github.com/golangci/golangci-lint/cmd/golangci-lint \
 				github.com/mna/pigeon \
 				github.com/modocache/gover \
 				github.com/onsi/ginkgo/ginkgo \
@@ -138,6 +137,7 @@ build: $(INSTALL_PREFIX) deps generate-optimized
 .PHONY: lint
 ## run golangci-lint against project
 lint:
+	@go get -v github.com/golangci/golangci-lint/cmd/golangci-lint
 	@golangci-lint run -E gofmt,golint,megacheck,misspell ./...
 
 
--- a/Makefile
+++ b/Makefile
@@ -65,14 +65,14 @@ help:/
           } \
         }' $(MAKEFILE_LIST)
 
-.PHONY: deps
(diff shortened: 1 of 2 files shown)


xcoulon:
@moorereason thanks for this PR! I guess that it also would make sense to call the new `linters` goal when running on Travis-ci: https://github.com/bytesparadise/libasciidoc/blob/master/.travis.yml#L36 (not on AppVeyor because I'm pretty sure the command will fail). Is it something you could include in your PR, please?

xcoulon:
Thanks for the PR update, @moorereason, but apparently, that does not work on [Travis-ci](https://travis-ci.org/bytesparadise/libasciidoc/jobs/442212246#L702): 
````
$ make linters
/bin/sh: 1: golint: not found
/bin/sh: 1: golint: not found
/bin/sh: 1: golint: not found
/bin/sh: 1: golint: not found
/bin/sh: 1: golint: not found
/bin/sh: 1: golint: not found
/bin/sh: 1: golint: not found
````
Also, I'm not sure if the linter will pass for all versions of Go defined in the build matrix, so we may want to only run the linter with the latest stable version of Go, for example.

**Comment by moorereason** [comment]
Pushed a new commit.  Feels a little hackish.  Do you want to maintain this?  I noticed after sending this PR that you have the GolangCI checker integrated into GH.  Is this PR necessary?  Open to suggestions

**Comment by xcoulon** [comment]
> Pushed a new commit. Feels a little hackish. Do you want to maintain this? I noticed after sending this PR that you have the GolangCI checker integrated into GH. Is this PR necessary? Open to suggestions

right, maybe providing a way to run [GolangCi linter](https://github.com/golangci/golangci-lint) on local would make more sense, indeed.

**Comment by moorereason** [comment]
That sounds like a good idea.  I forgot about GolangCI's cli interface.  I'll refactor.

File: Makefile [hunk 1]
```diff
@@ -58,12 +65,17 @@ help:/
           } \
         }' $(MAKEFILE_LIST)
 
-.PHONY: deps 
+.PHONY: install-devtools
+## Install development tools.
+install-devtools:
+	@go get -u -v $(DEVTOOLS)
```
xcoulon: I didn't know it's possible to pass a list of libs to install! That's awesome!

PR #233, opened 2018-11-21
fix(parser): fix parser failure on lists

### LINKED ISSUE #230: Avoid parser failure on pending docs in the 'compat' pkg
Some compatibility tests (in the `compat/fixtures/pending` pkg) fail with parsing errors, which is worse than not being able to recognize some elements of the asciidoc syntax.

xcoulon:
refactored the way mixed lists are built, using
a 'monitor' type to track changes of list item types
and call the appropriate list constructor with a buffer
of items, then "attach" the resulting list to its parent item

as a result, the 'pending' .adoc files can all be parsed
(even though the results remain different from the expectations,
but that's a different story)

fixes #230


pkg/types/grammar_types.go
```diff
@@ -925,18 +937,135 @@ func NewList(elements []interface{}) (List, error) {
...
+// NewList initializes a new `List` from the given content
+func NewList(items []interface{}) (List, error) {
+	log.Debugf("initializing a new List with %d items(s)", len(items))
+	monitor := newListMonitor()
+	for _, item := range items {
+		listItem, ok := toPtr(item).(ListItem)
+		if !ok {
+			return nil, errors.Errorf("item of type '%T' is not a valid list item", item)
+		}
+		monitor.process(listItem)
```
golangcibot: Error return value of `monitor.process` is not checked

PR #339, opened 2019-04-28
all: migrate from dep to Go modules

### LINKED ISSUE #325: Migrate from dep to go modules

shuLhan:
Steps when creating Go modules,
```
  $ go mod init github.com/bytesparadise/libasciidoc
  $ git rm Gopkg.lock Gopkg.toml
```
To make the package works with Go <1.11, we add the dependencies as
vendor,
```
  $ go mod vendor
```
Close #325


xcoulon:
@shuLhan thanks for this PR! One remark though: do we really need to include all the vendor content in the git repository? With `dep` there was no `vendor` content committed, and I'd rather keep is like this.

shuLhan:
I may have tried without adding vendor directory before but its failed when building on Go 1.10 on TravisCI.

I will push another commit (without vendor) and see how it go.

Thanks for reviewing.

xcoulon:
@shuLhan I'd like to make sure that the library builds on the 3 latest stable versions of Go, so I suggest we wait until Go 1.13 is released so we can remove the build on 1.10 and only rely on go modules from then on. This should probably simplify the build and remove the need for the `vendor` directory in the repo as well.

shuLhan:
> I'd like to make sure that the library builds on the 3 latest stable versions of Go, so I suggest we wait until Go 1.13 is released so we can remove the build on 1.10 and only rely on go modules from then on. This should probably simplify the build and remove the need for the vendor directory in the repo as well.

@xcoulon agree.  I try again last night without vendoring, and I think its work for 1.10.8 but the travis-ci seems stack.

xcoulon:
@shuLhan I removed the `tools.go` file from your PR. I can see that it references some libraries to make sure they are imported as modules, but it seems to duplicate the work done in the `make install-devtools` task. However, I'm fine with restoring if file if it helps in other places (maybe I missed something?). Please let me know

shuLhan:
@xcoulon from my understanding, the tools.go will allow Go tools to track the versioning of our tools [1].  If we don't have it, it will not listed in go.mod and go.sum.

[1] https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

xcoulon:
> @xcoulon from my understanding, the tools.go will allow Go tools to track the versioning of our tools [1]. If we don't have it, it will not listed in go.mod and go.sum. We can use "go get" in Makefile, but it will fetch the latest version in other user, maybe not the version that works and we use in our local.
> 
> [1] https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

ok, I see how it works now, and it looks actually better than forcing the version to install in the Makefile 👍 Thanks for the explanations, @shuLhan 🙌

PR #347, opened 2019-05-12
feat(parser/renderer): add user macro feature

### LINKED ISSUE #334: Custom tags definition
Is there a way to integrate custom tag definitions inside the parser?

odknt:
fixes #334 

* add parse rules for user macro to parser.
* add `renderer.DefineMacro()` that returns `renderer.Option` to renderer.
* user can define a rendering function for macro by call `renderer.DefineMacro()` and pass to `renderer.Wrap()`.

ex.

```go
func helloMacro(cm types.UserMacro) ([]byte, error) {
    // render content
}

...
    renderer.Wrap(context.Background(), doc, renderer.DefineMacro("hello", helloMacro))
...
```

### DIFF
--- a/pkg/parser/asciidoc-grammar.peg
+++ b/pkg/parser/asciidoc-grammar.peg
@@ -214,6 +214,7 @@ DocumentElement <- !EOF // when reaching EOF, do not try to parse a new document
             / DocumentAttributeDeclaration 
             / DocumentAttributeReset 
             / TableOfContentsMacro
+            / UserMacroBlock
             / Paragraph) {
     return element, nil
 }
@@ -549,6 +550,29 @@ TitleElement <- element:(Spaces / Dot / CrossReference / Passthrough / InlineIma
 TableOfContentsMacro <- "toc::[]" NEWLINE
 
 // ------------------------------------------
+// User Macro
+// ------------------------------------------
+UserMacroBlock <- name:(UserMacroName) "::" value:(UserMacroValue) attrs:(UserMacroAttributes) {
+    return types.NewUserMacroBlock(name.(string), value.(string), attrs.(types.ElementAttributes), string(c.text))
+}
+
+InlineUserMacro <- name:(UserMacroName) ":" value:(UserMacroValue) attrs:(UserMacroAttributes) {
+    return types.NewInlineUserMacro(name.(string), value.(string), attrs.(types.ElementAttributes), string(c.text))
+}
+
+UserMacroName <- (!URL_SCHEME !"." !":" !"[" !"]" !WS !EOL .)+ {
+    return string(c.text), nil
+}
+
+UserMacroValue <- (!":" !"[" !"]" !EOL .)* {
+    return string(c.text), nil
+}
+
+UserMacroAttributes <- "[" attrs:(GenericAttribute)* "]" {
+    return types.NewInlineAttributes(attrs.([]interface{}))
+}
+
+// ------------------------------------------
 // File inclusions
 // ------------------------------------------
 FileInclusion <- incl:("include::" path:(Location) inlineAttributes:(FileIncludeAttributes) { 
@@ -834,6 +858,7 @@ InlineElement <- !EOL !LineBreak
         / Link 
         / Passthrough 
         / InlineFootnote 
+        / InlineUserMacro 
         / Alphanums 
         / QuotedText 
         / CrossReference 
--- a/pkg/renderer/context.go
+++ b/pkg/renderer/context.go
@@ -2,11 +2,18 @@ package renderer
 
 import (
 	"context"
+	"errors"
+	"io"
 	"time"
 
 	"github.com/bytesparadise/libasciidoc/pkg/types"
 	log "github.com/sirupsen/logrus"
 )
+
+// MacroTemplate an interface of template for user macro.
+type MacroTemplate interface {
+	Execute(wr io.Writer, data interface{}) error
+}
 
 // Context is a custom implementation of the standard golang context.Context interface,
 // which carries the types.Document which is being processed
@@ -14,6 +21,7 @@ type Context struct {
 	context  context.Context
 	Document types.Document
 	options  map[string]interface{}
+	macros   map[string]MacroTemplate
 }
 
 // Wrap wraps the given `ctx` context into a new context which will contain the given `document` document.
@@ -22,6 +30,7 @@ func Wrap(ctx context.Context, document types.Document, options ...Option) *Cont
 		context:  ctx,
 		Document: document,
 		options:  make(map[string]interface{}),
+		macros:   make(map[string]MacroTemplate),
 	}
 	for _, option := range options {
 		option(result)
@@ -153,6 +162,15 @@ func (ctx *Context) GetImagesDir() string {
 		return imagesdir
 	}
 	return ""
+}
+
+// MacroTemplate finds and returns a user macro function by specified name.
+func (ctx *Context) MacroTemplate(name string) (MacroTemplate, error) {
+	macro, ok := ctx.macros[name]
+	if ok {
+		return macro, nil
+	}
+	return nil, errors.New("unknown user macro: " + name)
 }
 
 // -----------------------
--- a/pkg/renderer/html5/user_macro_test.go
+++ b/pkg/renderer/html5/user_macro_test.go
@@ -1,0 +1,135 @@
+package html5_test
+
+import (
+	"html"
+	texttemplate "text/template"
+
+	"github.com/bytesparadise/libasciidoc/pkg/renderer"
+	. "github.com/onsi/ginkgo"
+)
+
+var helloMacroTmpl *texttemplate.Template
+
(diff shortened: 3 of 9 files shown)


pkg/renderer/html5/user_macro_test.go
```diff
@@ -0,0 +1,116 @@
+package html5_test
+
+import (
+	"github.com/bytesparadise/libasciidoc/pkg/renderer"
+	"github.com/bytesparadise/libasciidoc/pkg/types"
+	. "github.com/onsi/ginkgo"
+)
+
+func helloMacro(cm types.UserMacro) ([]byte, error) {
+	name := cm.Value
```
xcoulon: you could use a buffer and templates here to render the content

odknt:
> I have not seen how those macros can be registered and associated with a template when running from the CLI (using a config file for example).

Sorry, I didn't read latest comment of the issue carefully.
I did refactor to be use Go template.

pkg/renderer/context.go
```diff
@@ -155,6 +164,15 @@ func (ctx *Context) GetImagesDir() string {
 	return ""
 }
 
+// MacroTemplate finds and returns a user macro function by specified name.
+func (ctx *Context) MacroTemplate(name string) (MacroTemplate, error) {
+	macro, ok := ctx.macros[name]
+	if ok {
+		return macro, nil
+	}
+	return nil, errors.New("Unknown user macro: " + name)
```
xcoulon: please avoid capital letters at the beginning of the error message. If the messages are wrapped/concatenated, we will have capital letters in the middle of the sentence.

xcoulon:
> how do you associate a template with a macro if you have multiple files in the `_macros` sub-directory?

ah, I see: the name of the file is the directly associated with the macro, isn't it? In this case, could we process the filename to remove its extension, so users can name their template files to something like `hello.tmpl` or `hello.txt`? 
Also, in the future, we may need to be able to configure the path to the directory which contains the templates. I'm not sure if `pkg/renderer/html5/_macros/` is suitable for users.

pkg/parser/asciidoc-grammar.peg
```diff
@@ -548,6 +549,29 @@ TitleElement <- element:(Spaces / Dot / CrossReference / Passthrough / InlineIma
 // ------------------------------------------
 TableOfContentsMacro <- "toc::[]" NEWLINE
 
+// ------------------------------------------
+// Custom Macro
+// ------------------------------------------
+UserMacroBlock <- name:(UserMacroName) "::" value:(UserMacroValue) attrs:(UserMacroAttributes) {
```
xcoulon: you may want to also capture the "raw text" in an extra argument when calling `types.NewUserMacroBlock(...)` so that if there is no such macro, the renderer can output the text as it was found in the `.adoc` source. For this, you can use `c.text` (`[]byte`) or `string(c.text)` and it will provide the entire text for the grammar rule that matched.

pkg/types/grammar_types.go
```diff
@@ -434,6 +434,44 @@ const (
 type TableOfContentsMacro struct {
 }
 
+// ------------------------------------------
+// Custom Macro
+// ------------------------------------------
+
+const (
+	// InlineMacro a inline user macro
+	InlineMacro UserMacroKind = "inline"
```
xcoulon: Good idea to use a `Kind` field to distinguish between inline and block macros. Would you mind you renaming this `UserMacroKind` type to just `MacroKind`, please, so I can use it to refactor the `InlineImage` and `BlockImage` types and merge them into a single `Image` type?

pkg/renderer/html5/user_macro_test.go
```diff
@@ -0,0 +1,118 @@
...
+	Context("user macros", func() {
+		opts := findMacroTemplate()
+
+		It("undefined macro", func() {
+
+			actualContent := "hello::[]"
+			verifyError(GinkgoT(), actualContent)
+		})
+
+		It("user macro block", func() {
```
xcoulon: should be `block macro` instead of `macro block`, I believe.

pkg/renderer/html5/user_macro_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+var _ = Describe("user macros", func() {
+
+	Context("user macros", func() {
+		opts := findMacroTemplate()
+
+		It("undefined macro", func() {
+
+			actualContent := "hello::[]"
+			verifyError(GinkgoT(), actualContent)
```
xcoulon: actually, if there is no such user-defined/custom `hello` macro, the output should be:

```
<div class="paragraph">
<p>hello::[]</p>
</div>
```

This is why I suggested that you also capture the "raw text" during the parsing, so it can be output in the HTML as-is.

pkg/renderer/html5/_macros/hello
```diff
@@ -0,0 +1,13 @@
+{{- if eq .Kind "block" -}}
+<div class="helloblock">
+<div class="content">
+{{end -}}
+<span>
+{{- if .Attributes.Has "prefix"}}{{escape (.Attributes.GetAsString "prefix")}} {{else}}hello {{end -}}
```
xcoulon: nice, I did not know about the use of parenthesis in the templates! thanks ;)

pkg/renderer/html5/_macros/hello
```diff
@@ -0,0 +1,13 @@
+{{- if eq .Kind "block" -}}
```
xcoulon: FYI, in https://github.com/bytesparadise/libasciidoc/issues/349 I've suggested that we have define the `inline` and `block` macros separately, as I believe that it's simpler to manage for the users (no need to change it in here for now, but that may be refactored in the future)

odknt: I think so too.

pkg/types/grammar_types.go
odknt: I will rename `UserMacroKind` with `MacroKind`, but I believe should be refactoring `InlineImage` and `BlockImage` in other PR.

pkg/renderer/html5/renderer_test.go
```diff
@@ -40,3 +34,35 @@ func verify(t GinkgoTInterface, expectedResult, content string, rendererOpts ...
 	diffs := dmp.DiffMain(result, expectedResult, true)
 	assert.Equal(t, expectedResult, result, dmp.DiffPrettyText(diffs))
 }
+
+func verifyError(t GinkgoTInterface, content string, rendererOpts ...renderer.Option) {
```
golangcibot: `verifyError` is unused (from `deadcode`)

pkg/renderer/html5/renderer_test.go
```diff
@@ -40,3 +34,35 @@ func verify(t GinkgoTInterface, expectedResult, content string, rendererOpts ...
...
+	require.NoError(t, err, "Error found while parsing the document")
+	t.Logf("actual document: `%s`", spew.Sdump(actualDocument))
+	rendererCtx := renderer.Wrap(context.Background(), actualDocument, rendererOpts...)
+	// insert tables of contents, preamble and process file inclusions
+	err = renderer.Prerender(rendererCtx)
+	require.NoError(t, err)
+	return rendererCtx
+}
+
+func newTextTemplate(name, src string, funcs ...texttemplate.FuncMap) texttemplate.Template {
```
golangcibot: `newTextTemplate` is unused (from `deadcode`)

pkg/types/grammar_types.go
xcoulon: yes, I'll take care of this in another PR, but having a generic name for the kind will make it easier, which is why I asked you if you could take care about it in your PR ;)

odknt:
Test failed on Windows due to LF replaced with CRLF in template file.
I thinking of remove `findMacroTemplate()` and replace template file with string.

Please tell me your thoughts.

xcoulon:
> Test failed on Windows due to LF replaced with CRLF in template file.
> I thinking of remove `findMacroTemplate()` and replace template file with string.
> 
> Please tell me your thoughts.

@odknt yes, it's safe to move the content of pkg/renderer/html5/_macros/hello into a string variable at this stage. We will have to deal with files in https://github.com/bytesparadise/libasciidoc/issues/349

One thing though: I had a similar problem a while ago, and I was able to fix it with the following git config: `core.autocrlf: input`. Maybe this will help?

pkg/renderer/html5/user_macro_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+import (
+	"html"
+	texttemplate "text/template"
+
+	"github.com/bytesparadise/libasciidoc/pkg/renderer"
+	. "github.com/onsi/ginkgo"
+)
+
+var macroOption renderer.Option
```
golangcibot: `macroOption` is unused (from `varcheck`)

odknt:
I squashed commits for cleanup.

odknt:
I added section about the feature to README.adoc, but I'm no good at English so my sentences may be terrible.

PR #370, opened 2019-06-30
chore(parser): add benchmark with real-world doc

xcoulon:
add the `verts-examples.adoc` file from
https://github.com/vert-x3/vertx-examples/blob/master/core-examples/README.adoc
and run the parser 50x

### DIFF
--- a/pkg/parser/bench_test.go
+++ b/pkg/parser/bench_test.go
@@ -1,6 +1,7 @@
 package parser_test
 
 import (
+	"io/ioutil"
 	"os"
 	"strings"
 
@@ -47,7 +48,6 @@ bar
 bar`
 			err := parseReader(source)
 			Expect(err).ShouldNot(HaveOccurred())
-			// b.RecordValue("expr count", float64(stats.ExprCnt))
 		})
 		timeout := 0.5
 		if ci {
@@ -64,7 +64,6 @@ bar`
 bar1`
 			err := parseReader(source)
 			Expect(err).ShouldNot(HaveOccurred())
-			// b.RecordValue("expr count", float64(stats.ExprCnt))
 		})
 		timeout := 0.1
 		if ci {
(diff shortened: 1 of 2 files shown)


moorereason:
My two cents:

I'm not a fan of Ginko's benchmarking because you're forced to give up the Go toolchain's benchmarking.  I think you'll have much better benchmark data if you just stick with the Go toolchain here.

To demonstrate, I refactored this commit to use regular Go benchmarking.  I then benchmarked master against a non-Sprintf pigeon (using your pending PR).  Comparing with [benchstat](https://godoc.org/golang.org/x/perf/cmd/benchstat), I get:

```
$ benchstat parser-bench-master.out parser-bench-nosprintf.out
name          old time/op    new time/op    delta
Parser10-2      1.19ms ± 3%    0.99ms ± 2%  -17.13%  (p=0.000 n=10+9)
Parser1-2        108µs ± 4%      98µs ±12%   -8.82%  (p=0.001 n=9+10)
ParserFile-2    98.9ms ± 7%    79.6ms ± 3%  -19.59%  (p=0.000 n=10+8)

name          old alloc/op   new alloc/op   delta
Parser10-2       325kB ± 0%     315kB ± 0%   -2.85%  (p=0.000 n=10+10)
Parser1-2       42.2kB ± 0%    41.6kB ± 0%   -1.40%  (p=0.000 n=9+9)
ParserFile-2    21.5MB ± 0%    20.5MB ± 0%   -4.70%  (p=0.000 n=9+10)

name          old allocs/op  new allocs/op  delta
Parser10-2       9.13k ± 0%     9.31k ± 0%   +1.95%  (p=0.000 n=10+10)
Parser1-2          716 ± 0%       732 ± 0%   +2.23%  (p=0.000 n=10+10)
ParserFile-2      786k ± 0%      795k ± 0%   +1.09%  (p=0.000 n=10+10)
```
Hope you find this useful.

**P.S.** - I'm running these tests on a VM on my laptop which can fluctuate some verses bare metal.

**P.P.S.** - The `go test` command I used was:
```
$ go test -run=XXX -bench=.* -benchmem -count=10 ./pkg/parser/ | tee parser-bench-nosprintf.out
```
The `-run=XXX` option is used to basically skip normal tests.

**P.P.P.S.** - I also benchmarked a `strings.Builder` version as suggested in the pigeon PR, but it didn't move the needle.
```
$ benchstat parser-bench-nosprintf.out parser-bench-builder.out
name          old time/op    new time/op    delta
Parser10-2       990µs ± 2%     998µs ± 4%    ~     (p=0.489 n=9+9)
Parser1-2       98.4µs ±12%    95.9µs ± 8%    ~     (p=0.211 n=10+9)
ParserFile-2    79.6ms ± 3%    78.4ms ± 1%  -1.45%  (p=0.011 n=8+9)

name          old alloc/op   new alloc/op   delta
Parser10-2       315kB ± 0%     315kB ± 0%    ~     (p=0.976 n=10+10)
Parser1-2       41.6kB ± 0%    41.6kB ± 0%    ~     (all equal)
ParserFile-2    20.5MB ± 0%    20.5MB ± 0%  +0.00%  (p=0.027 n=10+8)

name          old allocs/op  new allocs/op  delta
Parser10-2       9.31k ± 0%     9.31k ± 0%    ~     (all equal)
Parser1-2          732 ± 0%       732 ± 0%    ~     (all equal)
ParserFile-2      795k ± 0%      795k ± 0%    ~     (all equal)
```

xcoulon:
Thanks for the feedback @moorereason, this looks actually great! 🙌 

Would you like to open a PR with your refactoring, maybe? You could even reference the commands in the [README.adoc](https://github.com/bytesparadise/libasciidoc/blob/master/README.adoc) or even better in the [CONTRIBUTE.adoc](https://github.com/bytesparadise/libasciidoc/blob/master/CONTRIBUTE.adoc) so we can easily measure future changes/refactorings. WDYT?

PR #486, opened 2020-01-25
fix(parser): support comment between doc title and author

### LINKED ISSUE #481: // comment does not function properly if right after title
This is a difference in handling vs. asciidoctor

If I have 

= Title
// Copyright blah blah blah
// more boilerplate

Then the // is taken as an author.  If I insert a new line in front of the boilerplate, it's fine.

The main concern here is that this behavior differs from asciidoctor.  I have a large body of documents that lacks that extra whitespace.

xcoulon:
support comments (single lines and blocks) before the author
and between the author and the revision

Fixes #481


gdamore:
Looks good -- I'm thinking I may need to look more at this peg grammar support for other stuff. :-)

PR #582, opened 2020-06-06
Add support for (TM) and (R) in markup.

gdamore:
This adds trademark and registered trademark symbols.

I thought about adding support for the single and double arrows and emdash substitutions as well, but those are harder.

The arrows contain < and > and need to parsed and replaced before the &gt; and &lt; entities get substituted (so order of parsing is a problem).  The emdash actually needs a little context as well, as it only gets replaced when it is flanked appropriately, and it also can cause spaces to get replaced with thin spaces if it is flanked by spaces on both sides.

### DIFF
--- a/pkg/renderer/html5/string.go
+++ b/pkg/renderer/html5/string.go
@@ -21,7 +21,7 @@ func renderStringElement(ctx renderer.Context, str types.StringElement) ([]byte,
 	if err != nil {
 		return []byte{}, errors.Wrapf(err, "unable to render string")
 	}
-	result := convert(buf.String(), ellipsis, copyright)
+	result := convert(buf.String(), ellipsis, copyright, trademark, registered)
 	return []byte(result), nil
 }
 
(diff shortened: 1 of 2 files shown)


gdamore:
That was *quick* :-)

xcoulon:
ah, I forgot to check the commit message format 🤦‍♂️ 
please let's keep using the Commit message convention in https://gist.github.com/stephenparish/9941e89d80e2bc58a153

xcoulon:
> That was _quick_ :-)

I stumbled upon it while reading my mails with my morning coffee ;)

gdamore:
Ah, I didn't know about the commit header conventions.  In fact that document is the first time I've seen this particular commit comment conventions.

xcoulon:
well, I've been using it because there are tools that can generate a changelog based on the commit messages. But it's ok, I will manually add it to the changelog for the next release, and I'll be more careful in the future ;)

Issue #583, opened 2020-06-06
role not honored on h1 titles


gdamore:
The following markup:

```
[.myrole]
= Title
```

I would expect to result in setting a CSS class on the body.  (This is what asciidoctor-epub does, I've not checked asciidoctor HTML yet.)

For example:

```
<body class="article book-header">
....
</body>
```
I use this on front matter documents to allow my title page to be styled differently.

xcoulon:
> I use this on front matter documents to allow my title page to be styled differently.

I'm a bit confused with this sentence. Could you please elaborate on the "front matter documents" ?

xcoulon:
(also, I took the liberty to edit the description with fenced blocks so the Asciidoc and HTML outputs are visible)

gdamore:
No worries.  I use so many defect tracking systems, and they don't all use the same syntax for breaking out literal text....

For a book, you can have "front matter", which contains things like the title page, copyright page, dedication, and so forth.  Its often the case, especially for eBooks, that these need special styling or formatting, so its convenient to be able to use separate classes.

Although now that I think about it, maybe a simpler solution would be to just use totally different CSS files.

This would probably work because for various reasons I wind up doing  the front matter documents outside the spine (epub terminology again) -- so I basically just use a separate article style document (.adoc file) for each component.

xcoulon:
I just checked with Asciidoctor and the `role` attribute is supported, so let's have it here as well. 
Please let me know if you want to work on it, otherwise I can take care of it.

I believe that it's a matter of adding a `{{ if .Role}} {{ .Role }}{{ end }}` in https://github.com/bytesparadise/libasciidoc/blob/master/pkg/renderer/html5/html5.go#L35 and declaring the `Role` field in https://github.com/bytesparadise/libasciidoc/blob/master/pkg/renderer/html5/html5.go#L87. 
Then adding a test to make sure it works :)

gdamore:
Seems trivial enough, I'll go ahead and do it now just to clear it off the plate.

gdamore:
This turned out to not be quite so trivial, because the parser isn't picking up the Role.  I'm still figuring out that portion of the code.

PR #584, opened 2020-06-06
feat(renderer): role not honored on h1 titles (#583)

gdamore:
This adds support for setting the role on a level 0 title,
which will cause the class for the document body to be set.


gdamore:
The way I scraped the role off the attributes of the header felt a little slimy, but it seems the best way to match asciidoctor.

gdamore:
NB:  I fixed the style (apparently some whitespace crept in) -- please squash (or I can do that for you) before merging.

gdamore:
I squashed, because there was an unintended change (something I reverted) in the parser.

PR #586, opened 2020-06-06
refactor(types): avoid empty initializations

### LINKED ISSUE #573: Avoid empty initializations
Avoid initializing attributes, elements, etc. arrays when there's no content to store in them

xcoulon:
Avoid initializing element attributes, etc. until there's some actual content
use a `set()` func to initialize the Elementattributes on the fly when needed

Fixes #573

### DIFF
--- a/pkg/parser/blank_line_test.go
+++ b/pkg/parser/blank_line_test.go
@@ -16,7 +16,6 @@ second paragraph`
 		expected := types.DraftDocument{
 			Blocks: []interface{}{
 				types.Paragraph{
-					Attributes: types.ElementAttributes{},
 					Lines: [][]interface{}{
 						{
 							types.StringElement{Content: "first paragraph"},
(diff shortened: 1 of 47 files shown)


xcoulon:
I worked on this PR but I'm not 100% sure that the changes are worth it. 
The test data is simpler because there is no need for all those empty slides/maps, and there's a very small reduction in memory consumption, but that's at the cost of slightly more complex code to maintain, especially to avoid setting entries in nil maps! 🤔

xcoulon:
> I worked on this PR but I'm not 100% sure that the changes are worth it.
> The test data is simpler because there is no need for all those empty slides/maps, and there's a very small reduction in memory consumption, but that's at the cost of slightly more complex code to maintain, especially to avoid setting entries in nil maps! 🤔

code refactored in https://github.com/bytesparadise/libasciidoc/pull/586/commits/4f33ed47c64d1c1c895e10d2201d907f153f2783, using a new `Set(key, value)` method on the `ElementAttributes` which returns the resulting `ElementAttributes` which can be initialized on-the-fly if needed. 
This kinda makes this PR acceptable, now ;)

PR #589, opened 2020-06-06
Refactor bench and profile make targets

### LINKED ISSUE #580: Profiler in CI
Humble comments regarding the recent `bench` make target:

Why run the cpu/mem profiling in CI when you don't have an opportunity to view the resulting files?  It seems like a waste of CI compute time.

Additionally, running benchmarks in CI will not provide truly comparable CPU results (can't trust VM/container CPU resources/priority will be consistent between CI jobs).

My recommendations would be:

1. don't run benchmarking/profling in CI.
1. create a `bench` make target that just runs the benchmarks normally and saves the output to a temp file.  Run the target for each commit you want to compare and then use `benchstat` to compare the results.
1. move the profiling off to a separate make target.

I'm up for submitting a PR for these ideas if you're open to it.

moorereason:
Create a separate `profile` target for profiling. Only run the
benchmarks once in this target.

Update `bench` target to make five passes and save the output to a file
for use with `benchstat`.

Add a `bench-smoke` target to smoke test the top-level benchmarks by
running them only once.  This target is run via CI to ensure that the
benchmarks are functional.

Fixes #580

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -11,7 +11,7 @@ on:
 
 
 jobs:
-  
+
   test:
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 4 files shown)


xcoulon:
> Add a `bench-smoke` target to smoke test the top-level benchmarks by
> running them only once. This target is run via CI to ensure that the
> benchmarks are functional.

thanks, that's a nice alternative! I like the idea of running the bench but with a single iteration, just to make sure they work 👍

xcoulon:
(damn'it! second time today that I fail to check the commit message format 🤦‍♂️ )

xcoulon:
thanks for this PR @moorereason 🙌

moorereason:
I noticed after submitting that the commit subject didn't match you're standard, but you'd already merged it.  😄

xcoulon:
> I noticed after submitting that the commit subject didn't match you're standard, but you'd already merged it. 😄

yeah, I need to see if there's any GitHub App I could use to enforce that 🤔

PR #594, opened 2020-06-07
refactor(render) split out HTML5 templates

gdamore:
This breaks the former html5 renderer into a generic backend
SGML emitter, and an HTML5 specific flavor of that.  This
should facilitate other SGML-derived backends such as docbook,
XHTML, and semantic HTML.  It also lays the foundation for
other output formats like EPUB, markup-free ASCII, and perhaps
troff style output.

As part of this refactoring work, a bug related to the Context
being passed by value instead of by reference is fixed.

Also, some other minor style fixups, and improvements were made.
For example, bytes.NewBuffer(nil) was changed to &bytes.Buffer{}
(which potentially is also faster) and the use of type aliases
to improve readability.

Tests were unmodified except fixes needed to adjust for changes
to the package paths.


.goreleaser.yml
```diff
@@ -11,7 +11,7 @@ builds:
   # Path to main.go file or main package.
   # Default is `.`.
   main: ./cmd/libasciidoc
-  # Custom ldflags templates.
+  # Custom ldflags sgml.
```
gdamore, line 14: It looks like this was changed inadvertently.

gdamore:
cyclomatic complexity is a such a pain.  That function cannot really be made more simple.  I'll add an override.

gdamore:
Well, one way to do this would be to make prepareTemplates hard fail.   But I have another thought.  Stay tuned.

gdamore:
I'm going to squash this now, since it is passing.

gdamore:
This PR is so large, due to massive refactoring (moving files and splitting contents up) -- I recommend just checking the branch out and looking it over.    The stuff you want to be aware of:

The pkg/renderer/sgml package is not meant to be used directly, but provides the infrastructure for the pkg/renderer/sgml/html5 package to use. 

Templates are provided in the "Templates" object -- see pkg/renderer/sgml/html5/templates.go -- this is a big object.

The html5 package creates an sgml renderer by passing a default Templates object from it to the sgml.NewRenderer() method.

This means it is possible also to have multiple instances of these, with different templates.   However, at present the html5 package does not expose it's Templates object, nor does the sgml package offer a method to retrieve the templates it is using on it's Renderer.  It would be a trivial matter to expose those (via a new Templates() method or something like that), which would make customizing the templates programmatically a trivial matter -- one needn't provide a full set of templates but just override the templates as needed.

It's my thought that this could be used to provide a files-based renderer that simply loads the templates from a directory supplied, overriding only the templates which are supplied and using defaults otherwise.  That's future work, but I think makes making templates configurable a vastly simpler problem.

(One part that is sadly missing is good documentation about the templates.  At present one would need to have access to the code to figure this out.   As I think the details of these templates are still potentially in flux, it probably makes sense to wait to document too stringently.)

Oh, you probably also want to look at the code that sets up the function maps -- I've renamed some of the functions so that they are unique -- e.g. there is now "renderList" and "renderElements" as separate functions.  This allows us to just use a single global function map for all these.

gdamore:
I went ahead and followed up with the Templates() API I mentioned, and I also added code to only prepare the templates once, so that a given renderer can be reused multiple times without incurring the penalty of repeated reparses.

gdamore:
Ok, I added a test case for the Templates() API, which should satisfy coverage.  I hope.  I think I'm done with this PR.  I'm starting to make use of this for the XHTML backend.

gdamore:
I looked a bit at the codecov problem -- the problem is that the existing tests are inadequate.

Personally, I find the current test approach to be rather awkward -- ginkgo or whatever is most assuredly *not* easier than stock golang test.  What I'd have done is created input and output files in the test directory, and have a common function to check those. 

The randomization doesn't help either -- it made fixing problems a bit harder.  I also had no idea how to select just a single test case, since they are all hidden behind the ginkgo thing.

(For the records, I took a similar approach as you with mangos, except I used go convey.  In the end, I've found that trouble of the test framework wasn't worth the limited benefit we had in trying to get to an "English" style syntax.  I have been slowly converting hundreds of those test cases back to simple regular go test style test cases.  I've found the reduction in friction to be enormously helpful.)

Given a simpler testing framework, I'd probably endeavor to go ahead and add test cases -- in fact I'd like to do that for the new backend stuff I'm working on too, but I'm dreading having to interact with ginkgo and all that.

pkg/renderer/sgml/blank_line.go
```diff
@@ -0,0 +1,29 @@
+package sgml
```
xcoulon, line 1: would it be possible to make git aware that this file moved from `pkg/renderer/html5` to `pkg/renderer/sgml/html5`, so we don't loose its history? It seems like a handful of those files are deleted in the commit, but not all of them

xcoulon:
thanks for this PR, @gdamore! I will look into it

pkg/renderer/sgml/section.go
```diff
@@ -0,0 +1,89 @@
+package sgml
```
xcoulon, line 1: shouldn't this be part of the `html5` subpackage?

pkg/renderer/sgml/renderer.go
```diff
@@ -0,0 +1,360 @@
...
+	t, err = t.Parse(tmpl)
+	if err != nil {
+		log.Errorf("failed to initialize '%s' template: %v", name, err)
+		return nil, err
+	}
+	return t, nil
+}
+
+// Render renders the given document in HTML and writes the result in the given `writer`
+func (sr *sgmlRenderer) Render(ctx *Context, doc Document, output io.Writer) (Metadata, error) {
```
xcoulon, line 94: shouldn't this be part of the `html5` package? 
same question for all subsequent funcs/methods in this file

pkg/renderer/sgml/blank_line.go
gdamore: I tried to do that as much as possible, but it was mostly fixing up by doing some work -- originally I developed this by copying the entire tree and modifying.  Then to make the history right I did a backup, git mv, and copy over.  But it looks like I missed a couple.  It's kind of a pain, but if it's important to correct this for a few of those files, then I will see if I can go back and do it.

pkg/renderer/sgml/section.go
gdamore: I don't think so.  Why do you?

pkg/renderer/sgml/blank_line.go
xcoulon: yes, I'd like to keep the history clean. I believe it's doable by moving the files back to their older location, commit, then move again and commit again. If it's too much pain for you, I can take a look and contribute to your PR.

pkg/renderer/sgml/renderer.go
gdamore: I don't think so.

The strategy here was to make html5 just a collection of templates, and then a thin wrapper of instantiation of an instance of the sgml.Renderer with the given templates.

The purpose of this is to make it very easy for people to build "custom" backends by just supplying their own templates.  Basically consider the templates as configuration (but very powerful configuration because they can contain code), so this separates the configuration from the code.  I think all SGML derived documents are probably going to be able to use this code more or less as is, with only modification to the templates -- although I believe html5s and DocBook might need us to enhance the parser to separate semantic differences so that we can for example render <SAMP> vs. <KBD> instead of just <PRE> for both.  (This will mean providing two different templates, but for HTML5 and XHTML they may well be the same template.)

To be more concrete, the XHTML and ePUB backends will be able to use the SGML, and will probably not need to modify any code except to provide the same thin wrappers.  (Actually ePUB will need to do a fair bit more, but that's because it has to build a document structure in a ZIP file, but it's going to use the SGML backend to build up the XHTML assets that are included in the epub.)

Of course, if a given backend (such as PDF) is going to need more differentiation here, then they can provide their own Render implementation.  That's going to be a lot more work, so I'm hoping the cases for it will be rare.

pkg/renderer/sgml/section.go
xcoulon: ah never mind, I wasn't awake 🤦‍♂️

pkg/renderer/sgml/renderer.go
xcoulon: yeah, that's what we discussed earlier indeed. My question was wrong, as I was looking very briefly at the changes here. I'll give it a deeper look later today ;)

gdamore:
> >  I looked a bit at the codecov problem -- the problem is that the existing tests are inadequate.
>
> I'm not sure I understand what you mean by "inadequate" here. Could you elaborate?

The issue, if you dig in, is that the existing html5 tests don't exercise all the branches in the render. The neat thing about codecov is you can drill down and look at which lines of code are uncovered.  That can help inform the tests that you write. 

Using this approach, I was able to bring mangos (github.com/nanomsg/mangos) (which has a very much harder problem to solve for this, because it has to deal with network errors, and intrinsic race conditions, etc.) to 100% code coverage.  For a deterministic parser and generator, reaching 100% coverage should actually be relatively easy.

> For the record, so I started using Ginkgo because I thought it was a bit more easier and convenient. For example, the `-focus` argument takes the actual test name without having to convert to "snake format" as you said. There are also other features such as `BeforeEach`, etc, but I did not have to use them. I also wrote a few custom Gomega matchers to verify the parser and renderer output.
> 
> That being said, I'm not against converting all tests to the stock test framework, but it's going to be quite a big task I believe.

What I'd suggest is not forcing a conversion of all of them, but doing this in a "lazy" approach, so that you permit new tests to be written using the stock test framework.  Frankly, if I can just do that for the new backends, it will make creation of those backends easier for me.

xcoulon:
> > > I looked a bit at the codecov problem -- the problem is that the existing tests are inadequate.
> > 
> > 
> > I'm not sure I understand what you mean by "inadequate" here. Could you elaborate?
> 
> The issue, if you dig in, is that the existing html5 tests don't exercise all the branches in the render. The neat thing about codecov is you can drill down and look at which lines of code are uncovered. That can help inform the tests that you write.
> 
> Using this approach, I was able to bring mangos (github.com/nanomsg/mangos) (which has a very much harder problem to solve for this, because it has to deal with network errors, and intrinsic race conditions, etc.) to 100% code coverage. For a deterministic parser and generator, reaching 100% coverage should actually be relatively easy.
> 
ok, I'll need to check which branches I missed. But there's always the case of low-level errors which are harder to reproduce, though, so reaching 100% isn't going to be so easy.


> > For the record, so I started using Ginkgo because I thought it was a bit more easier and convenient. For example, the `-focus` argument takes the actual test name without having to convert to "snake format" as you said. There are also other features such as `BeforeEach`, etc, but I did not have to use them. I also wrote a few custom Gomega matchers to verify the parser and renderer output.
> > That being said, I'm not against converting all tests to the stock test framework, but it's going to be quite a big task I believe.
> 
> What I'd suggest is not forcing a conversion of all of them, but doing this in a "lazy" approach, so that you permit new tests to be written using the stock test framework. Frankly, if I can just do that for the new backends, it will make creation of those backends easier for me.

yes, that's an option, we just need to make sure that both can coexist and be executed on CI. But I think that it's doable

pkg/renderer/sgml/blank_line.go
gdamore: I tried a few times with this, and the combination of the rename and content change is enough for git decide that its effectively a deletion and addition.   To test I was to git mv, then I copied over file from another copy of the repo (using just cp, not git).  If you're able to figure out how to preserve the history, then please go ahead.  (You might be able to make this work for some of the files, but I couldn't make it work generally.)

xcoulon: ok, I'll give it a try then ;)

pkg/renderer/sgml/html5/templates.go
```diff
@@ -0,0 +1,84 @@
+package html5
+
+import (
+	"github.com/bytesparadise/libasciidoc/pkg/renderer/sgml"
+	"io"
+)
+
+var templates = &sgml.Templates{
```
xcoulon, line 8: cool, that's the kind of structure I was also thinking about. And it's even better here since it only contains the "string" templates (ie, not initialized at this point, that's something that can be done in the `sgml` package for all backends, indeed)

pkg/renderer/sgml/html5/templates.go
```diff
@@ -0,0 +1,84 @@
...
+	TocRoot:                 tocRootTmpl,
+	TocSection:              tocSectionTmpl,
+	UnorderedList:           unorderedListTmpl,
+	VerbatimLine:            verbatimLineTmpl,
+	VerseBlock:              verseBlockTmpl,
+	VerseBlockParagraph:     verseBlockParagraphTmpl,
+	VerseParagraph:          verseParagraphTmpl,
+}
+
+var defaultRenderer sgml.Renderer
```
xcoulon, line 67: I'm not sure we really need this `sgml.Renderer` interface. Or at least for now. I thought we could initialize the SGML renderer by passing it the `templates` as you did here, but only once and at the main package level, based on the backend chosen by the user.

pkg/renderer/sgml/renderer.go
```diff
@@ -0,0 +1,360 @@
...
+	// Templates returns the Templates used by this Renderer.
+	// It cannot be altered on a given Renderer, since the old
+	// templates may have already been parsed.
+	Templates() Templates
+}
+
+func NewRenderer(t *Templates) Renderer {
+	sr := &sgmlRenderer{
+		functions: make(funcMap),
+		templates: t, // TODO: should we copy this?
```
xcoulon, line 35: instead of copying here, you could avoid passing it as a pointer above.
By the way, I've tried passing as values rather than pointers as much as possible (I believe it's more efficient), unless it's absolutely needed (ie, the argument is going to be changed by the func).

pkg/renderer/sgml/renderer.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+	// Establish some default function handlers.
```
xcoulon, line 37: this could be set as 
```
sr := &sgmlRenderer{
	functions: map[string]interface{}{
		"render": sr.renderElements,
		...
	},
	templates: t,
}
```

pkg/renderer/sgml/renderer.go
```diff
@@ -0,0 +1,360 @@
...
+
+	"github.com/bytesparadise/libasciidoc/pkg/configuration"
+	"github.com/bytesparadise/libasciidoc/pkg/types"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+// Renderer implements the backend render interface by using sgml.
+type Renderer interface {
```
xcoulon, line 18: in order to keep the code as simple as possible, could we avoid introducing this interface for now? I understand it may be needed in the future, but only at this point should we have it, when we know exactly what we need.

pkg/renderer/sgml/templates.go
```diff
@@ -0,0 +1,196 @@
...
+	verseParagraph          *textTemplate
+}
+
+func (sr *sgmlRenderer) prepareTemplates() error {
+	t := sr.templates
+	var err error
+
+	sr.prepared.Do(func() {
+
+		sr.admonitionBlock, err = sr.newTemplate("admonition-block", t.AdmonitionBlock, err)
```
xcoulon, line 137: I like very much this way of propagating the error during the template setup! 👍 
it makes the code much simpler (not need to check for the error after each call to `newTemplate` 😓 )

pkg/renderer/sgml/templates.go
```diff
@@ -0,0 +1,196 @@
...
+	verseBlock              *textTemplate
+	verseBlockParagraph     *textTemplate
+	verseParagraph          *textTemplate
+}
+
+func (sr *sgmlRenderer) prepareTemplates() error {
+	t := sr.templates
+	var err error
+
+	sr.prepared.Do(func() {
```
xcoulon, line 135: in practice, we should only prepare the templates once. In which case did you need to protect that with a `sync.Once`? In the tests maybe?

pkg/renderer/sgml/blank_line.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"bytes"
+	"github.com/bytesparadise/libasciidoc/pkg/types"
+
+	log "github.com/sirupsen/logrus"
+)
+
+func (sr *sgmlRenderer) renderBlankLine(ctx *Context, _ types.BlankLine) ([]byte, error) {
```
xcoulon, line 10: I know it's a tiny thing, but could we use single letter for receiver names? Here, I believe that `r` (for "renderer") is enough. That's something I've tried to do in the rest of the code, so I'd like to keep things consistent here.

pkg/renderer/sgml/html5/templates.go
gdamore: I think it is potentially useful, as it allows for 3rd party applications to instantiate the HTML5 render and then just override certain templates, without having to subclass.

However, I guess you can also just copy the Templates and modify then call sgml.NewRenderer().

pkg/renderer/sgml/renderer.go
gdamore: So that's actually a fallacy, and IMO misguided.

Copy by value beats indirection in cases where you either are going to dereference every member anyway (so you're already copying anyway), or the object in question is small.

However, the total time difference is probably on the order of a few nanoseconds, and unless the dereferencing is done multiple times per member (it isn't here), it's likely a net wash (not even a few ns).

(For the record, I've spent a significant part of my life building very high performance software for use in circumstances like high frequency trading, IP protocol routing, etc.  At one point I was employed by Sun Microsystems as a contractor to optimize network performance to help them reach higher packet-per-second counts for IP forwarding.  A significant part of the effort was often "undoing" attempts to optimize code paths that ultimately lead to more complexity elsewhere, which turned out to result in costing the system more than just doing the simple thing.)

The bigger issue for me is that this approach smells like premature optimization.

Donald Knuth sums my opinion about this up pretty well:

“The real problem is that programmers have spent far too much time worrying about efficiency in the wrong places and at the wrong times; premature optimization is the root of all evil (or at least most of it) in programming.”

As a a concrete example here, pass by value optimization of the Context structure meant that the Table of Contents in the context was not updated properly.  It also led to another set of duplicate to create the Table of Contents (so we actually re-ran the steps to generate them!) and additional code. 

Generally, I think it's better to design the code for correctness first, and then optimize the big areas first (algorithm changes ,etc.)  Details like this count as micro-optimization, and in the vast majority are simply not worth the effort.  (The exceptions would be for cases where the code paths are *extremely* hot and time sensitive, such as low level context switching code, code taken for every IP packet, or improvements to bulk cryptographic algorithms.)

However, *in this case*, it might make sense to just copy  the structure (pass by value).  It's big, which is why I didn't do it.  The desire to copy would be to ensure that have our own value so that if someone wanted to later change the template we wouldn't be impacted (because we don't reparse the templates).

pkg/renderer/sgml/templates.go
gdamore: It shut up go cyclo, and reduced the total number of unexecuted code paths (for coverage) as well.

pkg/renderer/sgml/templates.go
gdamore: Render() calls this.  So this ensures that we actually only do this once.  This allows for initialization to be lazy, leading to simpler code elsewhere.

pkg/renderer/sgml/blank_line.go
gdamore: Funny, but I started out that way (it was originally t, because instead of sgmlRenderer it was templates, before I renamed the package.)

The problem was that as I was converting a lot of existing functions to move them into this case, there were a lot of name collisions.  That may or may not be the case with "r" or "s" on their own right now -- I haven't checked.  But "sr" is sufficiently unique that there are no collisions whatsoever.

I can try to do a rename of this mechanically with goland and see what happens.  If there are no collisions, I'm not adverse.

pkg/renderer/sgml/html5/templates.go
xcoulon: I would rather wait until we have other SGML renderers to see if such an interface helps, to be honest.

pkg/renderer/sgml/renderer.go
xcoulon: > So that's actually a fallacy, and IMO misguided.
> 
> Copy by value beats indirection in cases where you either are going to dereference every member anyway (so you're already copying anyway), or the object in question is small.
> 
> However, the total time difference is probably on the order of a few nanoseconds, and unless the dereferencing is done multiple times per member (it isn't here), it's likely a net wash (not even a few ns).
> 
Yes, that's my understanding as well. Also, the rule I've tried to follow so far was to avoid passing pointers if the argument does not need to be modified.


> (For the record, I've spent a significant part of my life building very high performance software for use in circumstances like high frequency trading, IP protocol routing, etc. At one point I was employed by Sun Microsystems as a contractor to optimize network performance to help them reach higher packet-per-second counts for IP forwarding. A significant part of the effort was often "undoing" attempts to optimize code paths that ultimately lead to more complexity elsewhere, which turned out to result in costing the system more than just doing the simple thing.)
> 

I believe you on this. I'm not sure there is more complexity elsewhere in the codebase here, but it's good to see that you have a lot of experience in this domain!

> The bigger issue for me is that this approach smells like premature optimization.
> 
> Donald Knuth sums my opinion about this up pretty well:
> 
> “The real problem is that programmers have spent far too much time worrying about efficiency in the wrong places and at the wrong times; premature optimization is the root of all evil (or at least most of it) in programming.”
> 

yes, I'm aware about this saying, and again, I wanted to follow to rule expressed above. My understanding is that the parser is by far the most time-consuming part of the whole process (by several orders of magnitude)

> As a a concrete example here, pass by value optimization of the Context structure meant that the Table of Contents in the context was not updated properly. It also led to another set of duplicate to create the Table of Contents (so we actually re-ran the steps to generate them!) and additional code.

I appreciate that you were able to find this bug and solve it :)

> 
> Generally, I think it's better to design the code for correctness first, and then optimize the big areas first (algorithm changes ,etc.) Details like this count as micro-optimization, and in the vast majority are simply not worth the effort. (The exceptions would be for cases where the code paths are _extremely_ hot and time sensitive, such as low level context switching code, code taken for every IP packet, or improvements to bulk cryptographic algorithms.)
> 
> However, _in this case_, it might make sense to just copy the structure (pass by value). It's big, which is why I didn't do it. The desire to copy would be to ensure that have our own value so that if someone wanted to later change the template we wouldn't be impacted (because we don't reparse the templates).

I'm sorry but this is where I'm lost: are you suggesting that in this case, passing by value is ok or not ok?

pkg/renderer/sgml/blank_line.go
xcoulon: ok, thanks!

pkg/renderer/sgml/templates.go
xcoulon: yes, I can imagine! That's great! 👍

pkg/renderer/sgml/renderer.go
gdamore: In this case, it may be better to do it.  I don't think it's a big deal either way.  I think pass by value may be better here because it describes *semantics* more accurately -- that is we really mean that future changes to the structure won't have any effect.

In other words, you convinced, but not because of performance.  :-)

pkg/renderer/sgml/html5/templates.go
gdamore: That's fine, I can remove this -- honestly this is mostly an artifact from earlier work on this code, and then I wound up creating the global Render() method and didn't need need the instances anymore.

My question still stands about whether you want me to wait for you to try fixing up the git history, or should I go ahead and fix the the items you've requested first?  I don't want to make the git history problem any worse if at all possible. :-)

pkg/renderer/sgml/renderer.go
gdamore: So looking at this, the html5 library needs to instantiate an object (an sgmlRenderer) with specific templates.  The xhtml5 library will do something similar.

So we need an instance to carry this.  At the same time, I would prefer to avoid exposing the details of the internals of the sgmlRenderer to consumers.  That way leads to API promises, and broken promises and broken hearts later, when we need to change things.

So I'd strongly prefer to keep this as an interface for now.

This can also help with writing test cases later, because stub implementations can be written to conform to the interface.  (For example, you might want to write a test Renderer that can be used to validate the document structure passed in by the parser.)

Interfaces are your friend, and one of the greatest aspects of go.  Don't fear them. :-)

pkg/renderer/sgml/html5/templates.go
gdamore: See my other comments -- I'm come back to my original position on this.  In fact, I am using this already in my xhtml5 work, and having the separate instances makes it possible to isolate state, permitting different formats to coexist in the same application.

So I vote *strongly* to keep this.  You'll almost certainly regret deciding not to later.

pkg/renderer/sgml/renderer.go
gdamore: Agreed.  Will make this change.

pkg/renderer/sgml/renderer.go
gdamore: I went ahead and made this change -- I'll commit it shortly.

pkg/renderer/sgml/renderer.go
gdamore: Can't *quite* do it that way, because I'm using  methods on the function being initialized.

But I *can* do:

sr.functions = funcMap{
   ...
}

So I'll do that.

pkg/renderer/sgml/blank_line.go
gdamore: So both "s" and "r" have conflicts.  (I tested both via mechanical renaming.)  I'd suggest to just let this one go.  We already have ctx anyway.

gdamore:
I've pushed an update as a single commit... (unsquashed).  This way you can either rebase my commit, or squash and do the git history fixups on your end.

pkg/renderer/sgml/html5/templates.go
xcoulon: Could you explain in which case you need to have separate instances (besides tests)? Why or when would you need to have different format coexisting in the same application?

pkg/renderer/sgml/renderer.go
xcoulon: For the record, I don't fear interfaces! I just don't want to introduce unnecessary artifacts that make the code more complex to maintain.

pkg/renderer/sgml/html5/templates.go
gdamore: Imagine a hugo instance.  It can emit XHTML, or HTML5, or ePUB based on for example an Accept-Content: header, or some other tag.  This is all in the same application server.

In my case, I have a tool chain that renders different document formats depending on command line switches.  (My tool chain generates HTML, ePUB, and PDF.  Right now I rely on being able to spawn separate instances of the process, but since I'm building a composite document out of literally hundreds of individual .adoc files, I'd really really like to do this in one shot.)

So for me, this is very much *not* contrived.

pkg/renderer/sgml/blank_line.go
xcoulon: ah too bad. I will probably give it a try then, in a more "manual" fashion. Are you ok if I make a PR on your PR so you can see what I want to change, rather than pushing on your PR directly?

pkg/renderer/sgml/templates.go
xcoulon: ok, fine

xcoulon:
> I've pushed an update as a single commit... (unsquashed). This way you can either rebase my commit, or squash and do the git history fixups on your end.

thanks, I've seen that. Yesterday evening I tried to fix the git commit history but so far I was not able to do it. I guess there are too many changes between the files for git to match them :(

gdamore:
> > I've pushed an update as a single commit... (unsquashed). This way you can either rebase my commit, or squash and do the git history fixups on your end.
> 
> thanks, I've seen that. Yesterday evening I tried to fix the git commit history but so far I was not able to do it. I guess there are too many changes between the files for git to match them :(

That matches my experience as well.

pkg/renderer/sgml/blank_line.go
gdamore: Sure, that's fine.

.goreleaser.yml
gdamore: I fixed that long ago.

pkg/renderer/sgml/blank_line.go
gdamore: Resolving this, as you tried and had the same results I did.

pkg/renderer/sgml/html5/templates.go
xcoulon: ok, so I'm not sure about the Hugo explanation, but I understand your use-case. Just to dig a little deeper on your toolchain: are you going to use the CLI or are you embedding the library elsewhere? because if you're using the CLI, then we'll probably need to support multiple backend values in the flags ;) (which is fine)

pkg/renderer/sgml/renderer.go
xcoulon: great, thanks!

pkg/renderer/sgml/html5/templates.go
gdamore: I am using my own go program.

I'm actually checking the doc sources out from github into a memory file system using a go implementation of git, and then doing the work there.  There's a fair bit of processing to do (links get rewritten, etc.) and getting this all in one program makes the whole conversion run in 1-2s, instead of several minutes.

xcoulon: > getting this all in one program makes the whole conversion run in 1-2s, instead of several minutes.

really?!? Wow, I'm impressed!

pkg/renderer/sgml/templates.go
```diff
@@ -0,0 +1,194 @@
...
+	VerbatimLine            string
+	VerseBlock              string
+	VerseBlockParagraph     string
+	VerseParagraph          string
+}
+
+type sgmlRenderer struct {
+	functions funcMap
+	templates Templates
+	prepared  sync.Once
```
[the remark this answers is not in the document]
xcoulon: I tried that but I could not get any good results (yet):
- firstly, if when we have multiple renderers, without the `sync.Once`, we'll probably endup initialize templates of all all renderers even if a single one is used, which is a waste of time. But maybe we could use a pkg variable next to the `defaultRenderer`?
- secondly, the pointer receiver is still needed on the `prepareTemplates()` func. But maybe we could refactor the code and make this method a pure func instead, which would return the initialized renderer? (just thinking at this point)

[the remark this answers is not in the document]
gdamore: pkg variables are not a good idea, I think.

Maybe we should just go with what we have?

pkg/renderer/sgml/aliases.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/bytesparadise/libasciidoc/pkg/renderer"
+	"github.com/bytesparadise/libasciidoc/pkg/types"
+)
+
+// These type aliases provide local names for names in other packages,
+// thereby helping minimize collisions based on conflicting package
+// names.  It also reduces the imports we have to use everywhere else.
+
+type Context = renderer.Context
```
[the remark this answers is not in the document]
xcoulon: > Using aliases means that I get to have less namespace pollution, which I think was part of why this was important to me. In particular, the renderer.Context was really annoying to me, because I couldn't use renderer as a symbol name anywhere else.

I see what you mean. I wished the alias could allow for a `renderer` type (not just a var): I tried renaming the `sgmlRenderer` type to `renderer` but it did not work :(

One reason why I don't like such aliases is that VSCode cannot "traverse" the alias and display the struct in the popup dialog when I hover the type

Eg:
image

vs 

image

> Basically the type alias like type Context = renderer.Context makes this symbol name a first class entity in my package (and exported), so there are no name collisions and lets me give a nice clean interface boundary to my package consumers.

sure, but it's a bit misleading IMO for the external developers, and anyways, the `pkg/renderer` is also public, so things should be clear IMO.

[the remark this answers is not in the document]
gdamore: The alias traversal not working in VSCode is unfortunate.  GoLand handles them beautifully.  I switch to GoLand from VSC a while ago, and I couldn't be happier with it.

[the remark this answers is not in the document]
xcoulon: > This kind of type aliasing is relatively new (go 1.11 or something like that).
> 
Yes, but IIRC, it was aimed at helping developers make a transition from one type to another. I wasn't aware of this use case.

> I've found it really helps to reduce the interface boundary between things. And narrow interfaces are something that I consider always a good design goal.
> 
> The other thing, is that having local aliases means that the code is far more readable -- at least to folks who aren't used to reading a lot of this,package.from.that.thing.but.imported.here.variable style of coding (Java).
> 
well, in go there's a single level of package, so I'm not sure what's the problem here?


> Meaning, to my mind, open(doc Document) is easier to read than open(doc types.Documents)
> 
to be honest I'm totally used to `open(doc types.Documents)`

> It's a small thing taken one at a time, but over time it gets to be ... annoying to type, and it disrupts the flow of the code.
> 

On the contrary, I believe it adds some mental load because I would always to remember that `Document` is an `Alias` for `types.Document` and if I want to navigate to the struct definition, I will have to go through the alias, first.

pkg/renderer/sgml/templates.go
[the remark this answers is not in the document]
xcoulon: well, package variables can be helpful, but not in this situation indeed. So yeah, let's stick with what you proposed.

xcoulon:
@gdamore this PR looks good to me like this and I believe we should go ahead and merge it so it does not block us in the subsequent tasks. Are you ok with this?

xcoulon:
@gdamore also, I created https://github.com/bytesparadise/libasciidoc/issues/596 so we have an issue to track this PR in https://github.com/bytesparadise/libasciidoc/projects/5

gdamore:
Yes, please, let's merge it.

xcoulon:
> Yes, please, let's merge it.

done! 🎉  
thanks a lot for this refactofing, @gdamore! 🙌

Issue #588, opened 2020-06-06
Support inline role assignment


xcoulon:
Eg:
```
[.rolename]`monospace text`
```

See https://asciidoctor.org/docs/user-manual/#inline-assignment-2

PR #600, opened 2020-06-14
599 Support for #mark 

### LINKED ISSUE #599: Support wanted for #mark# syntax
The #mark# syntax adds markup in one of two ways.

With the default role, it looks like highlighted text.  (It uses `<mark>`.)

With a role, it uses a `<span>` element with the class matching the mark.  This allows documents to apply an arbitrary style to a range of text, which they can then control via CSS.

The solution to this should build upon the work done by #588.

gdamore:
This is built upon my PR for #588, and depend on those changes.  It implements #mark support.

Reviewers should only look at the last PR, which will be rebased once #588 is merged.

This supports using the role to change the <mark> into a <span> with the class set appropriate, giving much flexibility to adding custom styling to regions of text.

This fixes #599

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -23,34 +23,34 @@ var g = &grammar{
 	rules: []*rule{
 		{
 			name: "AsciidocDocument",
-			pos:  position{line: 17, col: 1, offset: 332},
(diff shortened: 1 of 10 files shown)


gdamore:
I meant only look at the last commit in *this* PR.  The rest of the commits are the same as as #588.

PR #618, opened 2020-06-16
feat(renderer) Add xhtml5 support

gdamore:
This adds support for XHTML5 output, tests, and the -b
flag (which can now be used like -b [html,html5,xhtml,xhtml5].
(Also --backend.)  Tests are enclosed, and we leverage as much
as we can from the HTML5 backend.


pkg/renderer/sgml/xhtml5/templates.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/bytesparadise/libasciidoc/pkg/renderer"
+	"github.com/bytesparadise/libasciidoc/pkg/renderer/sgml"
+	"github.com/bytesparadise/libasciidoc/pkg/renderer/sgml/html5"
+	"github.com/bytesparadise/libasciidoc/pkg/types"
+)
+
+var templates = html5.Templates()
+
+//sgml.Templates{
```
gdamore, line 14: Ah I meant to delete this comment.  Will follow up.

xcoulon, line 14: could you remove this commented out lines?

xcoulon:
thanks for this contribution @gdamore 🙌

Issue #176, opened 2018-08-25
Render curved quotes


xcoulon:
See https://asciidoctor.org/docs/user-manual/#curved

shuLhan:
@xcoulon Can you give an example for this case?  I have an issues that may or may not related to single/double quotes, to minimize duplicate issue which I want to report.

Thank you.

xcoulon:
@shuLhan my case is that for text such as `mother's day` will be output as `mother&#8217;s day`. 

See https://asciidoctor.org/docs/user-manual/#curved. I'll edit the issue title and description to clarify.

xcoulon:
@shuLhan don't hesitate to open a new issue if you found a bug ;)

gdamore:
So it turns out that I need this too (I used this syntax extensively in my docs).

There are two dimensions of this problem.

First, we need to support "` and `"  (and also '` and `') elements.  These are inline elements like a String Element.  Notably asciidoctor does not process these "sanely" if they cross a paragraph (and presumably any block) boundary.  This probably creates some challenges to their use.  Which is unfortunate, because nobody really wants to type "&8217;."

Secondly, is the problem of automatically recognizing that a lone apostrophe should be curved when it appears without whitespace, along with the code to be smart about the use of "\" to escape them.

The particular rules also need to be savvy about `code` blocks, which creates some real challenges.  TBH, I believe that there can be ambiguity in the mark up language.

I'm not entirely sure how best to solve this.  The need is pressing for me, because I did use it heavily.  I might just "fix" my documents to use replacements instead.

I've got a solution which I think mostly does the right thing for apostrophes like mother's day, but a general solution to the explicit typographic entities is likely to be harder.

gdamore:
Actually in this day and age, we have unicode.  I know this is mean to be "ASCIIDOC", but really, there is no reason we can't inject UTF-8 explicit characters into place.  That's what I would recommend folks do for now.

gdamore:
See also #620.  I am considering *this* bug to just be the handling of mother's day style markup, with a renderer time substitution (for now -- although we might want to do this at parse time in the future).

gdamore:
So there are two cases to cover here.

1. mother's day (apostrophe with letters on both sides)
2. mothers`' day  (typographic curved apostrophe)

[...]

PR #621, opened 2020-06-18
feat(parser/render) Support typographic quote marks

### LINKED ISSUE #620: Support typographic curved quote marks
This is a request for "` and `", as well as the single quote variants '` and `'   

These should be treated as delimited blocks (at least that's how they work in asciidoctor), sort of like QuotedText elements.

These markups don't span paragraph blocks (empty lines).

Generally they get rendered with as unicode equivalents &#8216; &#8217; and such.

This issue is distinct from #176 which mostly is concerned with inline apostrophe marks.  The solution for #176 can be resolved by a post-markup substitution in the renderer like we do for other markup items (C) and such.

Instead, this one needs to handled by the parser, since it can be confused with monospace rendering otherwise.  (There are caveats with usage here... you generally will need to use unconstrained format monospace quoted text if adjacent to other elements.)

gdamore:
This adds support for the "`double`" and '`single`' types of
quoted strings.  These are inline enclosed elements, which is
necessary to prevent confusion with code blocks.  It also
matches asciidoctor.

Fixes #620

(PR NOTE: I did find a few bugs in the asciidoctor parser here; I have not attempted to be 100% bug compatible with asciidoctor.  This does conform to the documented, and what I believe, intended syntax was.  The PR is big mostly because of the large number of tests.  I will be following up with a fix for #176 once this PR integrates -- this PR is a bit more important to me than #176 as the fallback rendering for #176 is arguably still reasonable, but the fallback for this is not and quite ugly.)

### DIFF
--- a/pkg/parser/parser.peg
+++ b/pkg/parser/parser.peg
@@ -364,6 +364,81 @@ StandaloneAttributes <- attributes:(ElementAttribute)+ BlankLine* EOF { // stand
 }
 
 // ------------------------------------------
+// Quoted Strings (between curly single or double quotes)
+// ------------------------------------------
+
+QuotedString <- qs:(SingleQuotedString / DoubleQuotedString) {
+    return qs, nil
+}
+
+SingleQuotedString <- "'`" elements:SingleQuotedStringElements "`'" {
+    return types.QuotedString{ Kind: types.SingleQuote, Elements: elements.([]interface{})}, nil
+}
+
+SingleQuotedStringElements <- elements:(SingleQuotedStringElement+) {
+    return types.NewInlineElements(elements)
+}
+
+// We have to treat this one special, because of ambiguity with monospace markup.
+SingleQuotedStringElement <- element:(
+        LineBreak // must be before spaces
+        / Space+
+        / InlineIcon
+        / InlineImage
+        / InlineFootnote
+        / InlinePassthrough
+        / Link
+        / AttributeSubstitution
+        / BoldText
+        / ItalicText
+        / MarkedText
+        / SubscriptText
+        / SuperscriptText
+        / !"`'" MonospaceText
+        / DoubleQuotedString
+        / SingleQuotedStringFallbackCharacter) {
+    return element, nil
+}
+
+SingleQuotedStringFallbackCharacter <-  [^\r\n`] / "`" !"'" {
+    return types.NewStringElement(string(c.text))
+}
+
+DoubleQuotedString <- "\"`" elements:DoubleQuotedStringElements "`\"" {
+
+    return types.QuotedString{ Kind: types.DoubleQuote, Elements: elements.([]interface{})}, nil
+}
+
+DoubleQuotedStringElements <- elements:(DoubleQuotedStringElement+) {
+    return types.NewInlineElements(elements)
+}
+
+// We have to treat this one special, because of ambiguity with monospace markup.
+DoubleQuotedStringElement <- element:(
+        LineBreak // must be before spaces
+        // Space+
+        / InlineIcon
+        / InlineImage
+        / InlineFootnote
+        / InlinePassthrough
+        / Link
+        / AttributeSubstitution
+        / BoldText
+        / ItalicText
+        / MarkedText
+        / SubscriptText
+        / SuperscriptText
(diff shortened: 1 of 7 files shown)


pkg/parser/parser.peg
```diff
@@ -363,6 +363,81 @@ StandaloneAttributes <- attributes:(ElementAttribute)+ BlankLine* EOF { // stand
     return types.NewAttributes(attributes)
 }
 
+// ------------------------------------------
+// Quoted Strings (between curly single or double quotes)
+// ------------------------------------------
+
+QuotedString <- qs:(SingleQuotedString / DoubleQuotedString) {
```
xcoulon, line 370: you can just write it as 
```suggestion
QuotedString <- SingleQuotedString / DoubleQuotedString
```
ie, without capturing and returning the result in `qs`

pkg/parser/parser.peg
```diff
@@ -363,6 +363,81 @@ StandaloneAttributes <- attributes:(ElementAttribute)+ BlankLine* EOF { // stand
...
+    return qs, nil
+}
+
+SingleQuotedString <- "'`" elements:SingleQuotedStringElements "`'" {
+    return types.QuotedString{ Kind: types.SingleQuote, Elements: elements.([]interface{})}, nil
```
xcoulon, line 375: let's create a `types.NewQuotedString(type, elements)` func in `pkg/types/types.go` to be consistent with the rest of the code in this file

pkg/parser/parser.peg
```diff
@@ -363,6 +363,81 @@ StandaloneAttributes <- attributes:(ElementAttribute)+ BlankLine* EOF { // stand
...
+        / SingleQuotedStringFallbackCharacter) {
+    return element, nil
+}
+
+SingleQuotedStringFallbackCharacter <-  [^\r\n`] / "`" !"'" {
+    return types.NewStringElement(string(c.text))
+}
+
+DoubleQuotedString <- "\"`" elements:DoubleQuotedStringElements "`\"" {
+
```
xcoulon, line 408: could you remove this blank line?

pkg/parser/parser.peg
```diff
@@ -363,6 +363,81 @@ StandaloneAttributes <- attributes:(ElementAttribute)+ BlankLine* EOF { // stand
...
+    return types.QuotedString{ Kind: types.DoubleQuote, Elements: elements.([]interface{})}, nil
```
xcoulon, line 409: same here, let's add and use `types.NewQuotedString(...)`

pkg/parser/quoted_string_test.go
```diff
@@ -0,0 +1,1009 @@
...
+								},
+							},
+						},
+					},
+				},
+			}
+			Expect(ParseDraftDocument(source)).To(MatchDraftDocument(expected))
+		})
+		It("spaces with single quoted string", func() {
+			source := "'` curly was single `'"
```
xcoulon, line 35: do we allow spaces between the quotes and the text ?

gdamore: Yes.  This matches asciidoctor.

gdamore:
I've made the suggested changes, and verified the results.

gdamore:
Feel free to merge whenever, but be advised you should squash it when you do. :-)

pkg/parser/parser.peg
gdamore: Done.

pkg/parser/parser.peg
gdamore: Done.

pkg/parser/parser.peg
gdamore: Done.

pkg/parser/parser.peg
gdamore: Done.

pkg/parser/quoted_string_test.go
xcoulon: ok, good to know ;)

xcoulon:
> I've made the suggested changes, and verified the results.

thanks!

xcoulon:
thanks for this new contribution @gdamore 🙌

PR #624, opened 2020-06-19
fix(parser): quoted text newline handling

### LINKED ISSUE #623: Monospace quoted text incorrect spans paragraphs
I noticed that format like 

```
here is `some text

down` here
```

Winds up trying to create a `<code>` element that spans the blocks.  That's not conformant with asciidoctor, which just leaves the backticks where they are in the text.

Bold and monospace do not suffer this defect, it seems isolated to just code elements.

gdamore:
Quoted texts (such as bold) should span newlines, but should
not be permitted to span paragraphs or blocks.  This fixes
all of them, and includes parser and renderer validation.

Fixes #623

### DIFF
--- a/pkg/parser/quoted_text_test.go
+++ b/pkg/parser/quoted_text_test.go
@@ -77,6 +77,52 @@ var _ = Describe("quoted texts", func() {
 				Expect(ParseDraftDocument(source)).To(MatchDraftDocument(expected))
 			})
 
+			It("bold text with newline", func() {
+				source := "*some bold\ncontent*"
+				expected := types.DraftDocument{
+					Blocks: []interface{}{
+						types.Paragraph{
+							Lines: [][]interface{}{
+								{
+									types.QuotedText{
+										Kind: types.Bold,
+										Elements: []interface{}{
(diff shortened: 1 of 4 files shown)


gdamore:
I did not copy and modify the rendering tests for HTML to XHTML in this case.  I don't think they add any useful extra value in the XHTML tree, since the parse tree is already validated, and these formats are not different between HTML and XHTML.

pkg/parser/quoted_text_test.go
```diff
@@ -77,6 +77,52 @@ var _ = Describe("quoted texts", func() {
...
+				source := "*some bold\ncontent*"
+				expected := types.DraftDocument{
+					Blocks: []interface{}{
+						types.Paragraph{
+							Lines: [][]interface{}{
+								{
+									types.QuotedText{
+										Kind: types.Bold,
+										Elements: []interface{}{
+											types.StringElement{Content: "some bold\ncontent"},
```
xcoulon, line 90: now I realize that the `line` may actually span on multiple lines, but it's not a big deal I guess...

PR #632, opened 2020-06-20
refactor(parser): quoted text doesn't need !`\`

### LINKED ISSUE #631: quoted text grammar has useless ! rules
In the PEG grammar, ! can exclude things not seen (tokens not consumed yet), but it cannot look backwards.

Consequently rules that use this to exclude things in the beginning of a token simply don't work if those things might have already been consumed.

For example
```
DoubleQuoteMarkedText <- attrs:(QuotedTextAttrs)? !`\\` "##" elements:(DoubleQuoteMarkedTextElements) "##" { // double punctuation must be evaluated first
```

Could be expressed equally, and better, without the attempt to exclude the leading backslash - the rule for Escaped quoted text items takes precedence, if a leading backslash is seen.

This is basically true everywhere the !`\` rules are present for quoted text.

The grammar is complex enough -- let's simplify this.

gdamore:
This simplifies the grammar slightly.  May also improve the
performance.

Fixes #631

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -6698,31 +6698,22 @@ var g = &grammar{
 								},
 							},
 						},
-						&notExpr{
-							pos: position{line: 1033, col: 49, offset: 35971},
-							expr: &litMatcher{
-								pos:        position{line: 1033, col: 50, offset: 35972},
-								val:        "\\\\",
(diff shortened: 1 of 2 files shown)


xcoulon:
awesome! thanks @gdamore 🙌

PR #634, opened 2020-06-23
refactor(renderer): Simplify templates

### LINKED ISSUE #602: Support multiple roles on all blocks
A paragraph or a delimited block may have multiple roles (which should be retained in a slice).

Eg:
````
[.role1]
[.role2]
a paragraph
````

gdamore:
This eliminates the ContextualPipeline, and simplifies the templates
a bit, making them hopefully much easier to read and maintain, and
also to build derivatives.

The templates no longer require the use of functions, for the most
part. In particular the detailed rendering functions for nesting
are completely hidden from the templates, instead there is just a
generic .Content field which represents the rendered content for
nested content.

While here, I've added .Roles support to the top level blocks, although
I've not added test cases for them all.

A lot of the templates use the "sanitized" type to track the fact that
they can be trusted for HTML safety.

None of the tests were modified, in spite of the size of this change.

This is foundation work both for future backends (ePUB, docbook), as
well as further improvements to the existing rendering (for example
adding support for table styling.)

Fixes #602


xcoulon:
> While here, I've added .Roles support to the top level blocks, although
> I've not added test cases for them all.

ok, so could you add tests for that, or even better, move this feature in a separate PR if it's not too much work? I understand the refactoring you undertook and I like it (yes, the `ContextualPipeline` thing wasn't such a great idea), but finding changes in relation with support for multiple roles in the PR is a bit complicated...

gdamore:
It would be a pain to move the .Roles stuff out.  So I'd rather not do that.  It literally was like I was typing stuff, and didn't want to type .Role only to have to go back and fix it all later.  (No duplication of effort.)

I'll see about adding tests, but this is such a giant refactoring that I didn't want to wait for that to get it out for review.  I've spent *waay* too much time writing test cases (and fighting the test framework) -- the last couple PRs have been like 80% test code.

xcoulon:
> It would be a pain to move the .Roles stuff out. So I'd rather not do that. It literally was like I was typing stuff, and didn't want to type .Role only to have to go back and fix it all later. (No duplication of effort.)




> It would be a pain to move the .Roles stuff out. So I'd rather not do that. It literally was like I was typing stuff, and didn't want to type .Role only to have to go back and fix it all later. (No duplication of effort.)

I understand, but as I said before, having single focused PRs is much easier to review ;)

> 
> I'll see about adding tests, but this is such a giant refactoring that I didn't want to wait for that to get it out for review. I've spent _waay_ too much time writing test cases (and fighting the test framework) -- the last couple PRs have been like 80% test code.

Yes, I also spend most of my time writing tests, and that has been super useful in the past. But I still don't understand what are your pain points with the "test framework" exactly. Could you elaborate on that to see how we can make things easier? Is it just Ginkgo vs the standard testing package, or something else?

gdamore:
Actually. I don’t think I really fixed the roles problem because I’ve done nothing in the parser. This is pure renderer side stuff.

However what’s there means that when the parser is updated to populate the attributes properly then it will just work.

In the meantime we should leave the roles bug open.

And yes I have come to really hate ginkgo.   That’s not on you just a matter of taste.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #636, opened 2020-06-24
bug(parser): quoted strings should not allow interior adjacent spaces

### LINKED ISSUE #622: Smart quotes wrong when adjacent interior space present
I did some more conformance testing (was comparing other formats), and it appears that this format:

```
"` something `"
```

is wrong.   It looks like asciidoctor constrains the grammar so that interior adjacent whitespace to the rendered markup disqualifies them and they get rendered as the markup, instead of treated as smart quotes.

(I had specific test cases for this, but looking at asciidoctor it looks like I must have been confused about this.)

This is a bug, although reasonable people can disagree whether it is a bug in asciidoctor a bug in libasciidoc.  Still we should probably try to emulate the same behavior.

This appears to be limited to space characters.  Other characters like punctuation don't receive special treatment.

gdamore:
This doesn't catch all the ways that quoted strings should be treated
as constrained quoted text, but does fix this non-pathological test,
with fixed test cases.

Fixes #622

### DIFF
--- a/pkg/parser/parser.peg
+++ b/pkg/parser/parser.peg
@@ -369,7 +369,7 @@ StandaloneAttributes <- attributes:(ElementAttribute)+ BlankLine* EOF { // stand
 
 QuotedString <- SingleQuotedString / DoubleQuotedString
 
-SingleQuotedString <- "'`" elements:SingleQuotedStringElements "`'" {
+SingleQuotedString <- SingleQuoteStringStart elements:SingleQuotedStringElements SingleQuoteStringEnd {
     return types.NewQuotedString(types.SingleQuote, elements.([]interface{}))
 }
 
@@ -377,10 +377,14 @@ SingleQuotedStringElements <- elements:(SingleQuotedStringElement+) {
     return types.NewInlineElements(elements)
 }
 
+SingleQuoteStringStart <- "'`" ![ \t\r\n]
+
+SingleQuoteStringEnd <- "`'"
+
 // We have to treat this one special, because of ambiguity with monospace markup.
 SingleQuotedStringElement <- element:(
-        LineBreak // must be before spaces
-        / Space+
+        LineBreak !SingleQuoteStringEnd // must be before spaces
+        / Space+ !SingleQuoteStringEnd
         / InlineIcon
         / InlineImage
         / InlineFootnote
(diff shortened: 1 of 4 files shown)


pkg/parser/parser.peg
```diff
@@ -369,18 +369,22 @@ StandaloneAttributes <- attributes:(ElementAttribute)+ BlankLine* EOF { // stand
...
-SingleQuotedString <- "'`" elements:SingleQuotedStringElements "`'" {
+SingleQuotedString <- SingleQuoteStringStart elements:SingleQuotedStringElements SingleQuoteStringEnd {
     return types.NewQuotedString(types.SingleQuote, elements.([]interface{}))
 }
 
 SingleQuotedStringElements <- elements:(SingleQuotedStringElement+) {
     return types.NewInlineElements(elements)
 }
 
+SingleQuoteStringStart <- "'`" ![ \t\r\n]
```
xcoulon, line 380: you should be able to replace `![ \r\n\t]` with `[^ \r\n\t]`

pkg/parser/parser.peg
```diff
@@ -431,7 +435,11 @@ DoubleQuotedStringElement <- element:(
             return element, nil
 }
 
-DoubleQuotedStringFallbackCharacter <-  ([^\r\n`] / "`" !"\"") {
+DoubleQuoteStringStart <- "\"`" ![ \t\r\n]
```
xcoulon, line 438: same here

pkg/parser/parser.peg
gdamore: No I think that is wrong.  I don’t want to consume the token.  This leaves the next tokens available in the input stream without consuming them.

gdamore: This is important for nesting quotes text for example.

pkg/parser/parser.peg
gdamore: Same reply.

pkg/parser/parser.peg
xcoulon: ah right! I think that `&[^ \r\n\t]` would work in that case then. I like the `&` syntax because it really says that it looks forward in the content without consuming the next character.

gdamore: I think the two syntaxes are basically the same.  Using ! should convey that as well as & with one fewer characters.  I guess it’s all what you’re used to.

xcoulon: ok, I can try and refactor that later if you don't want to make the change here. It's not a big deal indeed.

gdamore: Actually there might be a difference - what happens with EOF.  Though that difference won’t matter in this particular case it might in others. 

More precisely the ! syntax says that the element is NOT followed by one of those characters whereas the & says that it IS followed by a character that is not in that list.

xcoulon: yes, I believe it's fine here, but it matters for delimited blocks for which the ending delimiter can be omitted when the block is at the end of the document. (Asciidoc is permissive on that regards.)

PR #642, opened 2020-06-25
bug(renderer): Wrong code for (TM)

### LINKED ISSUE #641: trademark character code "wrong"
Asciidoctor uses `&#8482;`  We used `&#153;`.  Both render identically (usually), but a little research shows that the higher order code is actually preferred.  We should also align for consistency with asciidoctor.  Some browsers handle the higher order one, but not the lower one.

gdamore:
We were using the code 153, which is not the same value used by
asciidoctor, and doesn't uniformly render as the trademark
symbol.

Fixes #641

(This PR should be mergeable/rebaseable as is, because it's changes are fairly isolated from other PRs currently outstanding.)

### DIFF
--- a/pkg/renderer/sgml/html5/string_test.go
+++ b/pkg/renderer/sgml/html5/string_test.go
@@ -28,7 +28,7 @@ var _ = Describe("strings", func() {
 	It("text with trademark", func() {
 		source := `TheRightThing(TM)`
 		expected := `<div class="paragraph">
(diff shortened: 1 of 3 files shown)


xcoulon:
good catch! thanks for this contribution @gdamore 🙌

gdamore:
Well, it was my fault in the first place. Lol.

PR #645, opened 2020-06-25
refactor(renderer): Use inline UTF-8 (optionally)

### LINKED ISSUE #626: Consider using inline UTF-8 and escaping as final pass in renderer
Right now we have a few places render special glyphs using unicode entities -- e.g. &#8217;

There is some inconsistency here, and I believe that there probably also some latent bugs in the treatment of these -- since it's possible for the rendering to be confused by what's meant at the parser step.

As we're moving toward using string internally (instead of []byte), I think we should just handle inline glyphs in their native UTF-8 until the final step in the rendering.

In fact, some backends have different rules, or are likely to, about what glyphs they want rendered and which they don't.  For example, the asciidoctor epub backend tends to just leave curved quotation marks in their unadulterated UTF-8, without converting them to HTML entities.  With modern HTML, this might even be preferable most of the time, as modern HTML5 requires UTF-8, so this rendering is going to be slightly faster for most users.

Probably we could use an "option" for renderers to specify whether they prefer to see non-ASCII characters turned into entities or not.

Note that entities are still required for a few special cases -- the `<`, `>`, `&`, and `"` symbols mostly.

gdamore:
This adds an option to render UTF-8 as is, doing the final
conversion to HTML entities only at the final step.  This
is more likely to result in correct encoding as they remain
unadulterated until the final rendering step.

Furthermore, some renderers may wish to allow Unicode to
be encoded as is.  This can be set on the document context,
but we also set it by default when a new "unicode" document
attribute is present.

Fixes #626


gdamore:
NB: I only added tests for this case to the HTML backend (not XHTML).  It shouldn't matter really, since the two backends share the same code in this regard.

pkg/renderer/sgml/html5/string_test.go
```diff
@@ -51,4 +51,16 @@ var _ = Describe("strings", func() {
 </div>`
 		Expect(RenderHTML(source)).To(MatchHTML(expected))
 	})
+
+	It("title with registered unicode", func() {
```
xcoulon, line 55: could you provide a test where the `unicode` attribute is not set (or set to `false`), so we can clearly see the difference in the output?

pkg/renderer/sgml/html5/quoted_string_test.go
```diff
@@ -26,6 +26,14 @@ var _ = Describe("quoted strings", func() {
 </div>`
 			Expect(RenderHTML(source)).To(MatchHTML(expected))
 		})
+		It("simple single quoted unicode", func() {
+			source := ":unicode:\n\n'`curly was single`'"
```
xcoulon, line 30: same question as below

pkg/renderer/sgml/html5/string_test.go
gdamore: The test just above this one has the same with it not set.

As with other boolean attributes, we don't have the notion of explicitly setting it to false.  Its enabled by its presence alone.

pkg/renderer/sgml/html5/quoted_string_test.go
gdamore: Same answer too... look just above -- I modified the test above by copy paste. :-)

xcoulon: yeah, that's what I thought just a few minutes ago 🤦‍♂️   
my apologies

PR #646, opened 2020-06-25
refactor(renderer): Clean up/simplify title handling

### LINKED ISSUE #640: Clean up/simplify title handling
Right now a bunch of the title processing is still passed through an explicit call to escape.

We should make renderElementTitle do the proper thing, and sanitize these things.

gdamore:
This eliminates the needs for templates to call escape.

Fixes #640


gdamore:
That was quick lol. :-p

PR #647, opened 2020-06-25
feature(renderer): htmlEscape should escape quotes

### LINKED ISSUE #644: htmlEscaper should escape quotes
Both ' and "" should be escaped for safety (after other replacements) as fields here may be placed into HTML attributes by templates.

For HTML (and SGML etc.) it is always safe to replace these with entities, provided they are part of user content (i.e. not explicitly part of the HTML markup itself.)

gdamore:
This makes all single and double quotes escaped in SGML content,
improving the safety of using these characters inline.

Fixes #644


gdamore:
This forced a lot of tests to be updated.

xcoulon:
great, thanks for taking care of this @gdamore!

PR #648, opened 2020-06-26
feature(parser/renderer): Render curved quotes

gdamore:
This provides support for both explicit typographical quotes, and
    the implicit curved apostrophe when it appears between letters
    as in mother's day.

    This also refactors the handling of certain typographic symbols,
    moving their handing from the render to the parser, so that they
    are treated properly during the parse phase.  (This was important
    in properly handling these curved quotes.)

    Also, the greedy rules for string elements in quoted text were
    removed, as they were interfering with the parse, and not strictly
    required.  (They were an optimization.)  It's our opinion that for
    most documents quoted text will be in the minority, such that
    the performance gain provided by the old rules was minimal at best.

This PR is based upon the work submitted for #644 (PR #647) and should be rebased once that PR is merged.

### DIFF
--- a/pkg/parser/parser.peg
+++ b/pkg/parser/parser.peg
@@ -387,6 +387,7 @@ SingleQuoteStringEnd <- "`'"
 SingleQuotedStringElement <- element:(
         LineBreak !SingleQuoteStringEnd // must be before spaces
         / Space+ !SingleQuoteStringEnd
+        / !"`" Symbol  // Exclude the explicit quote
         / InlineIcon
         / InlineImage
         / InlineFootnote
@@ -400,6 +401,7 @@ SingleQuotedStringElement <- element:(
         / SuperscriptText
         / !"`'" MonospaceText
         / DoubleQuotedString
+        / ImpliedApostrophe
         / SingleQuotedStringFallbackCharacter) {
     return element, nil
 }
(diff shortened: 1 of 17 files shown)


xcoulon:
@gdamore I just merged https://github.com/bytesparadise/libasciidoc/pull/647 so you can now rebase this PR and also fix the conflicts.

gdamore:
Rebased ... there were no conflicts just had to drop the first commit which you already merged. :-)

pkg/parser/parser.peg
```diff
@@ -1500,12 +1486,15 @@ UnquotedAttributeChar <- [^\r\n"=\],] { // excludes comma
     return types.NewStringElement(string(c.text))
 }
 
-// rule used after document attribute substitution, to discover links 
+// rule used after document attribute substitution, to discover links
+// TODO This is unused
```
xcoulon, line 1490: it's used in https://github.com/bytesparadise/libasciidoc/blob/master/pkg/parser/document_processing_apply_substitutions.go#L133, as the inline links can be "discovered" after substituting attributes

gdamore: Huh... I thought it came up unused in my editor... I'll remove it.

xcoulon: it's not used elsewhere in the grammar, but it is explicitly used as an entrypoint when parsing the line with substituted attributes (and the entrypoint is also defined in the `make generate` targets)

gdamore: Fixed, regenerated the parser, and rebased on top of master.

xcoulon:
thanks @gdamore 🙌

PR #649, opened 2020-06-26
chore(build): Use https://github.com/actions/cache

### LINKED ISSUE #591: Use https://github.com/actions/cache to speed-up builds
Since it supports Go modules, it may make the builds a bit faster

gdamore:
This uses https://github.com/actions/cache to speed up builds.

Fixes #591


xcoulon:
great! let's see how it goes, indeed ;)

xcoulon:
thanks @gdamore!

PR #652, opened 2020-06-26
chore(build): Use https://github.com/golangci/golangci-lint-action

### LINKED ISSUE #651: Use https://github.com/golangci/golangci-lint-action to speed up lint
The lint dependencies are not cached properly but there is an official GitHub action from the golangci-lint folks that is optimized for performance in GitHub.

Conceivably we could also enable this to work on Windows gracefully.

gdamore:
Fixes #651

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -9,16 +9,14 @@ on:
     branches:
       - master
 
-
 jobs:
-
   test:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         go-version: [1.13.x, 1.14.x]
         os: [ubuntu-latest, macos-latest, windows-latest]
-    name: ${{ matrix.os }} with Go ${{ matrix.go-version }}
+    name: Test ${{ matrix.os }} with Go ${{ matrix.go-version }}
 
     steps:
     - name: Install Go
@@ -41,11 +39,6 @@ jobs:
       run: |
         make build
 
-    - name: Lint
-      if: runner.os == 'Linux'
-      run: |
-        make lint
-
     - name: Test
       run: |
         make test-with-coverage
@@ -62,4 +55,19 @@ jobs:
       run: |
         make bench-smoke
 
+  golangci:
+    name: Lint with Go ${{ matrix.go-version }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go-version: [1.13.x, 1.14.x]
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Lint
+      uses: golangci/golangci-lint-action@v1
+      with:
+        version: v1.27
+        args: -E gofmt,golint,megacheck,misspell
 


gdamore:
This isn't right yet.

gdamore:
Note:  Per recommendations, we run the lint actions in their own job, to improve parallelism.

gdamore:
Git hub actions was confused by this commit, as I was changing workflow and it was busted for a bit.  Those 6 expected actions will never complete.

.github/workflows/ci-build.yml
```diff
@@ -9,16 +9,14 @@ on:
     branches:
       - master
 
-
 jobs:
-
-  test:
+   test:
```
xcoulon, line 13: the number of spaces (3) is weird. Would you mind fixing it?

gdamore: ok.  I didn't even realize this changed.

gdamore:
Should be ready to go now.

xcoulon:
> Should be ready to go now.

yes, I'm waiting for the jobs to complete and I'll merge it ;)

xcoulon:
there's something weird with the checks, though: there are 6 of them stuck with the following status `expected -- waiting for status to be reported`. Is it what you meant in https://github.com/bytesparadise/libasciidoc/pull/652#issuecomment-650424681 as well?

gdamore:
The six expected checks are confusion in GitHub actions... I effectively renamed some actions partway through this PR (added "Test" prefix), and that has caused GitHub some grief.

Part of this was probably also exacerbated by a syntax error in the YAML that caused the original tests to never finish, when I was working on it to begin with.  (Unfortunately in order to start this process and test it, I had to create a PR.  It wasn't right to begin with.)

If you want we can decline this PR, and I can start a new one which should be clean from the beginning (none of the screwy history breaking actions.)

xcoulon:
> The six expected checks are confusion in GitHub actions... I effectively renamed some actions partway through this PR (added "Test" prefix), and that has caused GitHub some grief.
> 
> Part of this was probably also exacerbated by a syntax error in the YAML that caused the original tests to never finish, when I was working on it to begin with. (Unfortunately in order to start this process and test it, I had to create a PR. It wasn't right to begin with.)
> 
> If you want we can decline this PR, and I can start a new one which should be clean from the beginning (none of the screwy history breaking actions.)

instead of declining it immediately, could you open a new PR with the commit you already have here to see if it works?

gdamore:
So I'm pretty sure that you have branch permissions set up to require certain status checks pass, and that this change breaks those settings.  I think you should review a bit more closely, because I'm sure you want to add the Lint rules, and you should add the Test prefix to the other rules.

xcoulon:
> So I'm pretty sure that you have branch permissions set up to require certain status checks pass, and that this change breaks those settings. I think you should review a bit more closely, because I'm sure you want to add the Lint rules, and you should add the Test prefix to the other rules.

oooh, indeed! My apologies, I totally forgot about those branch requirements! It's all good now, so I will merge this PR and close #654 if it's ok for you

PR #653, opened 2020-06-27
refactor(parser): Quoted Text Attribute refactoring

gdamore:
This refactors the handling of these to be more consistent,
in order to support future work to use more flexible (and
compliant with asciidoctor) attributes. This leads into the
fixes for making roles work everywhere, as well as providing
richer attributes for Table options.

### DIFF
--- a/pkg/parser/parser.peg
+++ b/pkg/parser/parser.peg
@@ -316,53 +316,70 @@ InlineAttributes <- "[" attrs:(GenericAttribute)* "]" {
     return types.NewAttributes(attrs.([]interface{}))								
 } 
 
-// Quoted Text attributes (attributes in front of marked up text).
-//
-// There are two forms, regular [role], and short-hand [.role#id].
-// They behave differently from each other and from other attribute lists.
-//
-// Document authors concerned with compatibility should stick to using
-// just alphanumerics, the hyphen, and the underscore, in these values,
-// and should also avoid whitespace.
-
-QuotedTextAttrs <- attrs:(QuotedTextRole / QuotedTextShortHand) {
-    return attrs, nil
+QuotedTextAttrs <- "[" attrs:(QuotedTextAttrRole? ShortHandAttr* NamedAttr*) "]" {
+    return types.NewElementAttributes(attrs)
 }
 
-// This is for [role].  Anything between the brackets is
-// considered, but it must not start with a # or dot.  Anything after a comma
-// (including the comma) is discarded.  (Future support for named attributes?)
-QuotedTextRole <- "[" role:QuotedTextRoleWord? "]" {
+QuotedTextAttrRole <- role:PositionalValue {
     return types.NewElementRole(role.(string))
-}
-
-QuotedTextRoleWord <- ![#.] role:([^\]]* {
-    return strings.SplitN(string(c.text), ",", 2)[0], nil
-}) {
-    return role, nil
-}
-
-// This is for [.role#id] (or [#id.role]).  Multiple roles can be present.  If multiple
-// ids are present only the first is retained.  Either the role or id may be omitted,
-// but at least one must be present.
-QuotedTextShortHand <- "[" attrs:( QuotedTextShortRole / QuotedTextShortID )+ "]" {
-    return attrs, nil
-}
-
-QuotedTextShortRole <- '.' role:([^.#\]]+ { 
-        return string(c.text), nil 
-    }) {
-    return types.NewElementRole(role.(string))
-}
-
-QuotedTextShortID <- '#' id:([^.#\]]+ {
-        return string(c.text), nil
-    }) {
-    return types.NewElementID(id.(string))
 }
 
 StandaloneAttributes <- attributes:(ElementAttribute)+ BlankLine* EOF { // standalone attributes, i.e., with nothing afterwards
     return types.NewAttributes(attributes)
+}
+
+ShortHandAttr <- ShortHandAttrID / ShortHandAttrOption / ShortHandAttrRole
+
+ShortHandAttrOption <- "%" option:ShortHandValue {
+    return types.NewElementOption(option.(string))
+}
+
+ShortHandAttrID <- "#" id:ShortHandValue {
+    return types.NewElementID(id.(string))
+}
+
+ShortHandAttrRole <- '.' role:ShortHandValue {
+    return types.NewElementRole(role.(string))
+}
+
+// PositionalValue is an unnamed attribute.
+PositionalValue <- ShortHandValue
+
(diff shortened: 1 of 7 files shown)


gdamore:
I will be following up with additional PRs to make use of this grammar for various block attributes.

Note that this also adds a new AttrOption handling, with support for %option stuff.  Those options are stored in the Attributes as a map[string]bool which should make it easier to use.  (Probably I'll also create a convenience wrapper like IsOptionSet(Attributes) -- follow up work.

pkg/parser/parser.peg
```diff
@@ -316,53 +316,82 @@ InlineAttributes <- "[" attrs:(GenericAttribute)* "]" {
...
-
-QuotedTextAttrs <- attrs:(QuotedTextRole / QuotedTextShortHand) {
-    return attrs, nil
-}
-
-// This is for [role].  Anything between the brackets is
-// considered, but it must not start with a # or dot.  Anything after a comma
-// (including the comma) is discarded.  (Future support for named attributes?)
-QuotedTextRole <- "[" role:QuotedTextRoleWord? "]" {
+// asciidoctor is inconsistent in handling quoted text attributes.
```
xcoulon, line 319: let's avoid such comments on Asciidoctor, please.

pkg/parser/parser.peg
```diff
@@ -316,53 +316,82 @@ InlineAttributes <- "[" attrs:(GenericAttribute)* "]" {
...
+NamedAttr <- ","? Space* key:NamedAttrKey Space* "=" Space* value:NamedAttrValue Space* {
+    return types.NewElementNamedAttr(key.(string), value.(string))
+}
+
+// The spec says attributes have be alphanumeric but does not consider foreign letters.  We are more generous.
+NamedAttrKey <- [\pL0-9_][\pL0-9_-]* {
+    return string(c.text), nil
+}
+
+// A few words about attribute values.  The official documentation says that they have to be quoted if they
```
xcoulon, line 368: what official documentation are you referring too? Asciidoc or Asciidoctor?

pkg/types/attributes.go
```diff
@@ -91,6 +91,10 @@ const (
 	AttrIconFlip = "flip"
 	// AttrUnicode local libasciidoc attribute to encode output as UTF-i instead of ASCII.
 	AttrUnicode = "unicode"
+	// AttrOptions element options (boolean, comma separated)
+	AttrOptions = "options"
+	// AttrOpts alias for AttrOptions
+	AttrOpts = "opts"
```
xcoulon, line 97: why do we need such an alias?

pkg/parser/parser.peg
gdamore: Ok. I wanted to relate that trying to match the "reference" implementation was .... challenging.

gdamore: Should I just delete the comment entirely?  Honestly you were probably the primary consumer for it -- although I guess if anyone else was surprised by the different handling for quoted text it might be good to note.

I've been trying to noodle through how asciidoctor handles attribute lists, and it *is* quite inconsistent.  There doesn't seem to be rhyme nor reason -- just different regular expressions I think.  Which, as I've noted, makes writing a reasonable grammar to emulate it's exact behavior somewhat .... challenging.  And ultimately I think not worth the effort -- the WG will probably clean up the grammar.

pkg/parser/parser.peg
gdamore: asciidoctor.

I am however following up on this -- because I wound up relexing this too -- we had too many tests that broke when I applied this (later commit) to more blocks.

I'll post an update here -- stay tuned.

pkg/parser/parser.peg
xcoulon: I understand your intent and I appreciate your effort in explaining the challenges that you encountered and the solution you're proposing. I just don't feel comfortable with having negative comment on Asciidoctor here, that's all. Maybe you could comment with something more neutral?

pkg/parser/parser.peg
gdamore: I do think I'll stop commenting on the grammar in comments though.  I generally would like to document the choices made, but it's almost impossible to do that without discussing asciidoctor, and without pointing out some of it's flaws where we differ.

I think it's probably better to have this discussion in the WG when we start working on a formal grammar.

pkg/types/attributes.go
gdamore: It's documented in asciidoctor.

pkg/parser/parser.peg
gdamore: The commentary on this PR will hopefully suffice for now.  I don't know how to point out that our behavior is different, without talking about the other implementation.  And frankly the other implementation's behavior is probably a limitation that is unintentional (i.e. bug).

pkg/parser/parser.peg
xcoulon: ok, sounds good to me. Let's stay neutral here but let's raise all those issues in the WG, then. But please don't feel offended by my request here!
Maybe we should start a doc somewhere to collect them as we keep working on the grammar? WDYT?

pkg/types/attributes.go
xcoulon: oh ok, I missed that.

pkg/parser/parser.peg
gdamore: Probably... I'm just not sure where to do that -- normally I'd think git would be the best place to do that (and even in the grammar spec itself), but I also understand your concerns.

The WG isn't going yet, they're still working on "governance", and I think it may be a while yet before it starts doing anything productive.

I just hope Dan & co. can refrain from adding "new features" (syntactic sugar especially) in the mean time.

pkg/types/attributes.go
gdamore: I've got a change (not here, will be follow up) that enables this grammar for all the blocks.  that use Attributes.  It doesn't yet add the support for this with [verse] and similar *yet*, but that's coming soon.  The main thing holding me back is writing all the *new* test cases.

(I also had to fix some tests, since adding the new syntax rules broke some of the tests that were actually incorrect and reflected implementation rather than documented syntax.)

gdamore:
Please see the follow up commit here-- I'll squash for you, but I want you to be able to see the delta first.

xcoulon:
> Please see the follow up commit here-- I'll squash for you, but I want you to be able to see the delta first.

ok, looks good

gdamore:
And squashed.

gdamore:
Argh.  One sec.

gdamore:
Fixed now. .. mismerge (missing curly bracket doh!)

xcoulon:
thanks @gdamore 🙌

PR #654, opened 2020-06-28
chore(build): Use https://github.com/golangci/golangci-lint-action

### LINKED ISSUE #651: Use https://github.com/golangci/golangci-lint-action to speed up lint

gdamore:
Fixes #651

(This is a resubmit of 651, which should be clean.)

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -9,16 +9,14 @@ on:
     branches:
       - master
 
-
 jobs:
-
   test:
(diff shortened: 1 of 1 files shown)


gdamore:
Well it wasn't.  Do you have those other test cases somewhere in github settings?  I don't have access to that.

PR #656, opened 2020-06-28
refactor(parser/renderer): fail when file to include is not found

### LINKED ISSUE #655: fail when a file to include cannot be found
Instead of reporting the error and rendering the error in the HTML output (which can be hard to see), the document preprocessor should fail. 
Having a missing file to include is a severe error, so rendering an incomplete doc (or page) does not make sense, IMO.

xcoulon:
Changing the behaviour when the preprocessor cannot find the file to include
(file missing or invalid path).
Instead of producing a draft document
with an embedded error message
which can be tedious to locate/notice,
the preprocessor fails (hard).

Also, no more error logging in the console

Fixes #655


xcoulon:
The parser changed even though I did not touch the grammar for this PR. I suspect this is a platform issue. @gdamore what OS are you running on?

PR #664, opened 2020-06-29
refactor(parser): VerseBlock is redundant

gdamore:
The code for quote blocks already switches to verse block if
the [verse] attribute is present, thus there is no need for
extra parsing steps to attempt to distinguish verses.  This
reduces complexity, and should provide a small performanc boost.

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -1431,6 +1431,9 @@ func NewDelimitedBlock(kind BlockKind, elements []interface{}, attributes interf
 		log.Debugf("overriding kind '%s' to '%s'", kind, k)
 		kind = k
 	}
+	if k, ok := attrs.GetAsString(AttrStyle); ok {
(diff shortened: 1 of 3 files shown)


pkg/types/types.go
```diff
@@ -1431,6 +1431,9 @@ func NewDelimitedBlock(kind BlockKind, elements []interface{}, attributes interf
 		log.Debugf("overriding kind '%s' to '%s'", kind, k)
 		kind = k
 	}
+	if k, ok := attrs.GetAsString(AttrStyle); ok {
```
gdamore, line 1434: This will be used as I convert the delimited blocks over to the new BlockAttrs syntax -- it should reduce the transition effort.

xcoulon:
Nice! It annoyed me that I had to maintain verse vs quote blocks, so I'm glad you found a way to simplify this!

PR #666, opened 2020-06-29
feature(renderer): Support reversed lists

### LINKED ISSUE #662: Ordered lists should support %reversed option
To be honest, I'm not sure what the use case for the %reversed option is.  But asciidoctor supports it, so we should probably add it.

Once the refactoring for ordered list is integrated, adding it is probably relatively straight-forward.  (Has anyone seen this used in the wild?)

gdamore:
Fixes #662


gdamore:
I'll come with this fixed properly.  I must have missed something in my testing.

PR #668, opened 2020-06-29
feature(renderer): Support reversed lists

### LINKED ISSUE #662: Ordered lists should support %reversed option

gdamore:
While testing, found some more potential edge cases around
the parser for attributes lists, so tightened those up.

Fixes #662

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -2690,31 +2690,51 @@ var g = &grammar{
 								name: "ShortHandValue",
 							},
 						},
+						&andExpr{
+							pos: position{line: 344, col: 50, offset: 11847},
+							expr: &charClassMatcher{
+								pos:        position{line: 344, col: 51, offset: 11848},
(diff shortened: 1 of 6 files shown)


xcoulon:
@gdamore looks good but there is a couple of formatting issues reported by the linter

gdamore:
was extra lines (whitespace) in the test file.  (My editor stopped formatting on save for some reason...)

Fixed and squashed.  Should be ready to go.

xcoulon:
> was extra lines (whitespace) in the test file. (My editor stopped formatting on save for some reason...)
> 
> Fixed and squashed. Should be ready to go.

thanks!

PR #669, opened 2020-06-30
feature(parser/render): Multiple roles for images.

gdamore:
This also includes bug fixes for parser.  We've tried to
emulate most, but not all, of the quirks found in asciidoctor
with respect to images.

(Following commentary not in the git log:)

This PR refactors (and removes) some of the logic intended to make verseblocks and quoteblocks work like asciidoctor.  Apparently asciidoctor treats [quote] differently from [verse] when process image blocks.  Both of these cases are actually pathological, since imageblocks should never be the content of a verse or a quote (the result was actually undesirable as well -- leaving source markup in the output).  In order to bring some sanity, we treat them both the same, and assume that any attributes ahead of an image:: block really are meant to apply to the image.

We also noted that the handling for inline attributes vs. block attributes is inconsistent, in strange ways.  Both use the positional parameter as the "alt" for the image, but inline does not permit shorthand role or ID elements (I think this was noticed as a bug that I filed against asciidoctor), and the 2nd and 3rd positional parameters for height and width are only honored in the inline attributes (in block attributes they are ignored.)

Trying to find some sanity here, we opted to consistently treat the 2nd and 3rd parameters as width and height, regardless of where they are.  We do treat the alt tag (first positional parameter) differently as that seemed like it might break existing documents if we did otherwise (though documents that rely on that might be considered defective).  Note that as a consequence of this, it is impossible to apply multiple roles to an inline image.  (We would encourage fixing this, and making inline attributes more strictly follow the block attribute list rules -- that can be a follow up PR.)

There is some strangeness around the acceptance or not of whitespace in attribute lists.  The grammar fixes here are meant to make us conform more tightly to what asciidoctor does, but again it's not 100%. The places where differ are very pathological -- and easily worked around.

### DIFF
--- a/pkg/parser/parser.peg
+++ b/pkg/parser/parser.peg
@@ -41,13 +41,13 @@ AsciidocDocumentBlocks <- header:(DocumentHeader?) others:(DocumentBlock*) {
 
 DocumentBlock <- 
         LabeledListItem // must appear before simple paragraph
+        / ImageBlock
         / SimpleParagraph
         / BlankLine // must be before LiteralBlock 
         / Section
         / DelimitedBlock
         / FileInclusion
-        / VerseParagraph // must be before ImageBlock
-        / ImageBlock
+        / VerseParagraph
         / ThematicBreak
         / OrderedListItem
         / UnorderedListItem 
@@ -64,11 +64,12 @@ DocumentBlock <-
 AsciidocDocumentBlocksWithinDelimitedBlock <- blocks:(DocumentBlockWithinDelimitedBlock)*
 
 DocumentBlockWithinDelimitedBlock <- !EOF 
-    block:(DelimitedBlock
+    block:(
+        ImageBlock
+        / DelimitedBlock
         / BlankLine // must be before LiteralBlock
         / FileInclusion
         / VerseParagraph
-        / ImageBlock
         / ThematicBreak
         / OrderedListItem
         / UnorderedListItem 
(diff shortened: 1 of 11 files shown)


pkg/parser/parser.peg
```diff
@@ -1543,18 +1561,47 @@ ResolvedExternalLink <- url:(ResolvedLocation) inlineAttributes:(LinkAttributes)
...
 
-InlineImage <- "image:" !":" path:(Location) inlineAttributes:(ImageAttributes) {
-    return types.NewInlineImage(path.(types.Location), inlineAttributes.(types.Attributes))
+ImageAltAttrInline <- value:InlinePositionalValue {
+    return types.NewElementNamedAttr(types.AttrImageAlt, value.(string))
 }
 
-ImageAttributes <- "[" alt:(StandaloneAttributeValue)? ","? width:(StandaloneAttributeValue)? ","? height:(StandaloneAttributeValue)? ","? Space* otherattrs:(GenericAttribute)* "]" {
-    return types.NewImageAttributes(alt, width, height, otherattrs.([]interface{}))
+ImageAltAttr <- Space* value:PositionalValue Space* {
```
xcoulon, line 1586: I assume that the support for spaces around the positional value in the `ImageAltAttr` rule vs lack of support for spaces in the `ImageAltAttrInline` rule is related to the comment below found in the PR?

> There is some strangeness around the acceptance or not of whitespace in attribute lists.

gdamore: You would assume correctly.  I am trying hard not to describe all the ways in which asciidoctor lacks consistency.

xcoulon: yes, that one more area in which I hope the WG will clarify!

xcoulon:
thanks a lot for all the investigation and all the work in this PR, @gdamore! 🙌

PR #670, opened 2020-06-30
Namespace

gdamore:
This has one commit on top of the imageblock fixes.  It changes the value for AttrCustomID and AttrPositional2/3 to avoid using names that users could use (it uses '@' character to isolate the namespace).


xcoulon:
good idea, this will prevent accidental overrides 👍

PR #672, opened 2020-06-30
refactor(parser/renderer): Support section and document roles

gdamore:
This adds support for both sections and the top-level document
(section0 elements) to have multiple roles.  It also adds the
support for the ID element for documents, in a way that precisely
emulates what asciidoctor does (only custom IDs are emitted).


xcoulon:
thanks @gdamore!

PR #673, opened 2020-06-30
chore(docs): Update Readme and limitations docs

### LINKED ISSUE #612: Update Known Limitations file for all gaps in parity with Asciidoctor
It seems that there are more features missing for parity with Asciidoctor than what is listed on this page.  https://github.com/bytesparadise/libasciidoc/blob/master/LIMITATIONS.adoc

This issue is to provide a case to reference for a PR to update the limitations document.

gdamore:
This adds a lot more detail covering what is supported and
what is not to the documentation files.

There might be missing items still, but I've tried to cover
everything we know about.

Fixes #612


xcoulon:
awesome! this was something I neglected, so thanks you having taken time to update these docs! 🙌

PR #683, opened 2020-07-02
refactor(parser): Minor refactoring of icon attributes

gdamore:
This changes NewNamedElementAttributes to be a varargs style
function, simplifying use, and allows us to eliminate the
special function used for alloction icon attributes.

While here, removed an unused parser rule.


xcoulon:
thanks for this refactoring @gdamore!

PR #684, opened 2020-07-03
bug(parser): Inline attribute parsing is incorrect

gdamore:
(Note that block attributes, and some other inline attributes,
suffer from this problem -- this PR only fixes icon and image
inline attributes.)

While working on some new support for align and float attributes,
we noticed that the parser was fragile, and failed on certain
cases including double embedded whitespace, embedded quote marks,
and others.

This logic will be expanded to other elements and to block
attributes; this incremental PR should be easier to review.

### DIFF
--- a/pkg/parser/attributes_test.go
+++ b/pkg/parser/attributes_test.go
@@ -1,0 +1,255 @@
+package parser_test
+
+import (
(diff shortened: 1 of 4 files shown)


xcoulon:
thanks @gdamore!

Issue #676, opened 2020-06-30
Support alignment and float attributes for images


gdamore:
This is useful for tightly controlling document layout, and likely used frequently in documents containing inline figures or diagrams.

gdamore:
This should mostly be just picking up things in the SGML renderer and putting these fields into the templates -- the details should be managed via CSS.

gdamore:
This was actually fixed with the integration of #685

PR #685, opened 2020-07-04
feature(renderer): Add image float, align, and capture support.

gdamore:
Note that the global figure-caption override is still not honored.

(Text below is not in the git history).

I have changes to support figure-caption, but they require adding new default attributes at document root.  This will require updating ever test in the parser suite -- > 350 when I looked -- to add these stock attributes.  As this is going to be such a large job, I'd rather not do it multiple times as we add other global caption attributes (e.g. for TIP, CAUTION, and Example blocks, etc.)   Instead I'd like to do those later, all at once.   Unless someone has a better idea how to deal with "unsetting" default attributes (i.e. the main use of figure-caption is in a disable -- :figure-caption!:  block.

### DIFF
--- a/LIMITATIONS.adoc
+++ b/LIMITATIONS.adoc
@@ -88,7 +88,8 @@ Interactive checklists are not supported.
 
 Interactive SVG support is missing, as is support for inline SVG.
 
-Float, alignment, and caption attributes have no effect.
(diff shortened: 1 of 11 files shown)


gdamore:
Fixes #676.  Oh, and I should have said "caption" not "capture".  Lol.

PR #688, opened 2020-07-04
refactor(parser): default subs on paragraphs and delimited blocks

### LINKED ISSUE #633: default substitutions on blocks and paragraphs
Prior to working on #558 and #597 which address support for custom substitutions on delimited blocks and paragraphs, some refactoring on the parser needs to be undertaken, so we can use the default `normal` and `verbatim` substitutions.

xcoulon:
Changing the way delimited blocks and paragraphs (including continued
paragraphs within lists, but NOT list item contents) are pre-processed:
- the draft document now contain the "raw" content for these paragraphs
and delimited blocks, using the new `types.RawLine` type
- in order to "build" the final document, a preprocessing phase is
undertaken, during which the default substitutions are applied on the
raw contents.

In other words, substitutions are resolved in 2 phases of parsing.

Also:
- using the new "MatchInlineElements" to check inline elements
- paragraphs lines are now `[]interface{}` instead of `[][]interface{}`,
which allows for containing a singular `types.RawLine` after the first
parsing phase.

Fixes #633


xcoulon:
@gdamore FYI.

I have to solve the latest merge conflicts, but I'm tempted to make a few more changes in the substitutions, so they would be applied when generating the "draft document", which would result in something more consistent. What bothers me in this current PR is that the draft document can be a mix of:
- paragraphs and delimited blocks with "raw content"
- other blocks (sections, list items, etc.) with fully-parsed content.
So, I will probably revisit the PR so that the draft doc looks closer to what we had before. Oh well, looks like I'll need to update a lot of tests again 😬

gdamore:
So while you're working on this, we should be aware that attribute values that are single quoted are supposed to be subject to substitutions as well.  They aren't today.  Also attribute values are supposed to be rendered as they may contain markup -- right now we do that expansion later in the document -- but we might want t think about doing it earlier.  To facilitate this it might be helpful to have a different kind of StringElement()  -- sort of an InlineMarkupElement that contains inline content that has not been processed yet.

I'm not proposing any of this for right now -- just wanting to plant these thoughts in your brain so you can keep them in mind as you do the work.

xcoulon:
> So while you're working on this, we should be aware that attribute values that are single quoted are supposed to be subject to substitutions as well. They aren't today. Also attribute values are supposed to be rendered as they may contain markup -- right now we do that expansion later in the document -- but we might want t think about doing it earlier. To facilitate this it might be helpful to have a different kind of StringElement() -- sort of an InlineMarkupElement that contains inline content that has not been processed yet.
> 
> I'm not proposing any of this for right now -- just wanting to plant these thoughts in your brain so you can keep them in mind as you do the work.

yes, so right now I'm thinking about having the following types of documents:
- `RawDocument`: resulting from the first parsing phase, with file inclusions resolved. This, it would contain "raw" elements in paragraphs and delimited blocks
- `DraftDocument`: what's in `master` branch now: based on the `RawDocument`, with all substitutions resolved (ie, on paragraphs and delimited blocks, and also on attributes). 
- `FinalDocument` (or simply `Document`): where sections and lists are "reorganized" in a hierarchical manner, and preamble and table of contents marker are included.

As a consequence, the attributes substitutions you mentioned could be part of the processing of the `RawDocument` which produces the `DraftDocument`.

PR #690, opened 2020-07-04
feature(renderer): Top level table styling

gdamore:
This provides top-level support for attributes to style the
entire table (frame, grid, float, stripes, width, autowidth).

The cols attribute is still not handled, meaning that it is
still not possible to configure column widths.

### DIFF
--- a/pkg/types/attributes.go
+++ b/pkg/types/attributes.go
@@ -71,16 +71,10 @@ const (
 	AttrLastUpdated = "LastUpdated"
 	// AttrImageAlt the image `alt` attribute
 	AttrImageAlt = "alt"
-	// AttrImageWidth the image `width` attribute
-	AttrImageWidth = "width"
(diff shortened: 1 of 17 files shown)


gdamore:
I also renamed some attributes (the Go types, not the string values) to make better reuse of them for tables.

gdamore:
(Whoops, forgot to regenerate the parser...)

pkg/types/attributes.go
```diff
@@ -99,6 +93,16 @@ const (
 	AttrCaption = "caption"
 	// AttrStyle block or list style
 	AttrStyle = "style"
+	// AttrWidth the `width` attribute used ior images, tables, and so forth
+	AttrWidth = "width"
```
xcoulon, line 97: Maybe we should also rename `AttrImageHeight` to `AttrHeight` for the sake of consistency?

PR #698, opened 2020-07-05
features(types/renderer): Table cols attribute support

### LINKED ISSUE #686: Limitations list doesn't reflect the state of affairs with quoted text handling
I've just tested the tool and https://github.com/bytesparadise/libasciidoc/blob/master/LIMITATIONS.adoc#quoted-text seems out of date to me: both examples in that section work exactly as in Asciidoctor.

By the way, very good tool, congrats!

### LINKED ISSUE #694: The style attribute should be supported on table columns
This is the remaining cols attribute handling needed, and when combined with #693 will provide complete handling for the cols attribute.

Implementing this will likely require changing the parser to provide contextual parsing of the table cells.

gdamore:
This adds support for handling the cols attribute ont tables,
giving the ability specify per column widths and alignments.
It also supports the repeat syntax, and the ability to set
autoexpand (either for the entire table or for an individual row).

The width handling and calculations were moved to instantiation
time for the Table, so that all backends can benefit from the
calculation work done for these.

The style component of the column specification is consumed and
stored, but does not affect deeper parsing or rendering yet.

This takes table support about as far as it can go without
reworking the parser to provide context-sensitive parsing of
the table contents.

While here the limitations document was updated to provide more
detail on limitations, including links to issues for folks
who want to track progress.

Fixes #694
Fixes #686


gdamore:
Full rebased now.

I believe this PR *substantially* increases the number of tables in the wild that will be handled correctly.

xcoulon:
also, thanks for the big update in `LIMITATIONS.adoc` 👍

PR #701, opened 2020-07-06
refactor(renderer) Newlines in should be in templates, not renderer

### LINKED ISSUE #699: Newlines in should be in templates, not renderer
The renderer for SGML injects newlines between elements at various times. This is unfortunate, as it leads to extra complexity in the renderer, and leaves a portion of the generated output outside of the control of the templates.

Elements (block elements) that should have a newline appended should have that reflected in the templates, rather than having the SGML core assume this. 

I noticed this while working on an epub backend.

gdamore:
This refactors the renderer somewhat so that newlines are no
longer injected in between block elements by the renderer,
but rather the closing block element has a newline appended
in the closing block tag in the template.

This moves the control over these newlines to the templates,
and makes the logic to understand when they get added easier
to understand.

This affects pretty much all the tests that look for HTML output,
as they now have a newline at the end (unless they are enclosed
within full html tags.)  For the most part, they just had a newline
appended their expected output.

Note that this does not introduce stray whitespace in the *middle*
of documents.

Two places where stray newlines were introduced (possibly for
compatibility with asciidoctor) was chagned -- if the document body
is empty then we do not emit a blank line.  Likewise we do not do
so if a blockquote has no body content.

This will not have any effect on rendered output.

Fixes #699


gdamore:
NB: I don't know why asciidoctor had those stray blank lines -- I think that was an implementation artifact and not intentional.  It certainly made the rendered output look kind of strange in the editor (why is this blank line here?)

gdamore:
Note also: this will make my epub templates "cleaner" -- and easier to follow without resulting in weird intervening whitespace.

xcoulon:
thanks @gdamore!

PR #711, opened 2020-07-11
feature(renderer): version-label not honored

### LINKED ISSUE #710: version-label not honored
The version label is meant to be something that can be overridden, or reset, e.g. via

:version-label!:

or

:version-label: Something Else

Note that asciidoctor does something really bizarre here, as it lower cases the supplied version label, and does not support quoting the value.  I consider that a defect in asciidoctor.

gdamore:
This supports the version-label be changed, or even reset
for document attributes.

To support resetting this label, the initial value is located
in the Predefined attributes, and a reset of the attribute is
stored as a nil attribute value instead of deleting it from the
attributes.  This allows us to distinguish a reset as an override.

This approach will be used for other predefined text values in
follow ups.

Fixes #710

### DIFF
--- a/pkg/types/attributes.go
+++ b/pkg/types/attributes.go
@@ -109,6 +109,8 @@ const (
 	AttrPositional2 = "@2"
 	// AttrPositional3 positional parameter 3
 	AttrPositional3 = "@3"
+	// AttrVersionLabel labels the version number in the document
+	AttrVersionLabel = "version-label"
 )
 
 // NewElementID initializes a new attribute map with a single entry for the ID using the given value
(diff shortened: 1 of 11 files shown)


pkg/types/attributes.go
```diff
@@ -338,32 +340,38 @@ func (a Attributes) NilSafeSet(key string, value interface{}) {
...
+		if value == nil {
+			return "", false // nil here means attribute was reset
+		}
 		if value, ok := value.(string); ok {
 			return value, true
 		} else if v, ok := a[key]; ok {
 			return fmt.Sprintf("%v", v), true
 		}
 	}
+	// check in predefined attributes
```
xcoulon, line 353: I guess that storing a `nil` value instead of removing it is because you also changed the priority for looking up attributes? We don't want to look in the predefined attributes if there was an explicit reset beforehand?

gdamore: Yes.  I need a way to distinguish something that was *reset* vs something that was simply not assigned.

xcoulon: yes, ok, it makes sense.

PR #715, opened 2020-07-11
feature(types/parser/renderer): Support for counters

### LINKED ISSUE #714: Want attribute substitutions for {counter:name} and {counter2:name}
The {counter:xxx} counter format of attribute substitution is supposed to let us access counters.

This is likely to be useful for users, and also can help when defining attributes for overrides of captions.

gdamore:
This adds support for {counter:name} and {counter2:name}
as attribute substitutions that reference the named counters.

The existing table, image, etc. counters are not converted to use
these yet, but it's likely that doing so will help once attribute
substitution is supported.

Note that when that work is done, care must be taken to ensure that
the counter is incremented when expanded to it's literal value, not
just expanded from the built-in value.

Fixes #714


pkg/parser/document_processing_apply_substitutions.go
```diff
@@ -74,6 +76,37 @@ func applyAttributeSubstitutions(elements []interface{}, attrs types.AttributesW
 
 }
 
+func applyCounterSubstitution(c types.CounterSubstitution, attrs types.AttributesWithOverrides) (interface{}, bool, error) {
+	counter := attrs.Counters[c.Name]
```
xcoulon, line 80: should you do  something like 
```
counter, found := attrs.Counters[c.Name]
if !found {
    counter = 0
}
```

pkg/parser/parser.peg
```diff
@@ -202,10 +202,34 @@ AttributeReset <- ":!" name:(AttributeName) ":" Space* EOL {
     return types.NewAttributeReset(name.(string))
 }
 
-AttributeSubstitution <- "{" name:(AttributeName) "}" {
+xAttributeSubstitution <- "{" name:(AttributeName) "}" {
```
xcoulon, line 205: is it meant to be removed?

pkg/parser/paragraph_test.go
```diff
@@ -167,6 +167,22 @@ foo`
 				}
 				Expect(ParseDraftDocument(source)).To(MatchDraftDocument(expected))
 			})
+
+			It("with counters", func() {
+				source := `foo{counter:foo} bar{counter2:foo} baz{counter:foo} bob{counter:bob}`
```
xcoulon, line 172: the Asciidoctor docs says that `counter2` can be used to increment the counter without displaying it: https://asciidoctor.org/docs/user-manual/#counters

Also, when processing the source above, Asciidoctor produces the following result:
```
<div class="paragraph">
<p>foo1 bar baz3 bob1</p>
</div>
```

pkg/parser/parser.peg
gdamore: yes. that was a goof.

pkg/parser/paragraph_test.go
gdamore: Whoa.  I  totally misread the counter2 docs -- I though it was display without incrementing, not the reverse.  Will fix.

gdamore: I will come back with a revised PR with that fixed.

I recommend that we integrate the fix for the stray debugging message (you can cherry-pick or I can submit a separate PR for that), and then I will squash this all together along with the fixes.  (It is not convenient to keep them separate, as I have to refactor the CounterSubstitution a bit to fix the revised understanding of counter2.)

gdamore:
Actually you've picked up the stray debug message.  Stay tuned for an update.  I will probably squash this.

gdamore:
Please *re-review*.  I believe I've fixed the semantics of counter2.

I've also added support for using the set syntax with counter2 (to set the starting index without displaying) in precisely the same manner that asciidoctor does.  (Tested this against asciidoctor as well.)

pkg/parser/document_processing_apply_substitutions.go
gdamore: The default value for the counter will be zero anyway.  So that's fine.

pkg/parser/document_processing_apply_substitutions.go
```diff
@@ -74,6 +76,48 @@ func applyAttributeSubstitutions(elements []interface{}, attrs types.AttributesW
 
 }
 
+func applyCounterSubstitution(c types.CounterSubstitution, attrs types.AttributesWithOverrides) (interface{}, bool, error) {
```
xcoulon, line 79: could you add some comments on this func to explain the returned values. Also, Do you really need to return a `bool`? AFAICS, it's always `true` unless there is an error.

pkg/parser/document_processing_apply_substitutions.go
gdamore: Well, default will be *nil* is what I meant to say.  I just didn't see a reason to introduce another variable.

pkg/parser/document_processing_apply_substitutions.go
gdamore: I can.  I think I used the same context for bool as ApplySubsitutions did -- I wanted this to conform to the same signature.

gdamore: Basically it has to conform to applyAttributeSubstitutionsOnElement.

gdamore: Comment applied -- I will squash if you approve.

xcoulon: ok, I see. Thanks for the clarification.

pkg/parser/document_processing_apply_substitutions.go
xcoulon: sure, but my point was that `counter, found := attrs.Counters[c.Name]` is more idiomatic (I believe)

gdamore: I don't think that's necessarily more idiomatic -- unless you want to capture the case where a nil value might have been stored in the map vs. not being present in the table at all.

I think the idiomatic approach is to have zero values be reasonable.

Having said that, this is mostly down to a matter of taste.  If you want me to make the change I will.

xcoulon:
> Please _re-review_. I believe I've fixed the semantics of counter2.
> 
> I've also added support for using the set syntax with counter2 (to set the starting index without displaying) in precisely the same manner that asciidoctor does. (Tested this against asciidoctor as well.)

cool, thanks for fixing the PR. I'll merge it once all builds passed.

gdamore:
Want me to rebase this?

xcoulon:
> Want me to rebase this?

I've already rebased the PR and I can squash the commits when merging. Unless you want to do it yourself?

gdamore:
Nah, if you're good handling it, that's 100% fine with me. ;-)

xcoulon:
PR merged. Thanks @gdamore! 🙌

Issue #604, opened 2020-06-14
Support attribute substitution in attribute values


xcoulon:
For example:
```
:markup-in-source: verbatim,quotes

[source,java,subs="{markup-in-source}"]
----
System.out.println("Hello *bold* text").
----
```

PR #717, opened 2020-07-12
feature(renderer): Tables should honor table-caption attribute

### LINKED ISSUE #712: Tables should honor table-caption attribute
The table-caption document attribute should be honored, including supporting being reset.

gdamore:
This uses a kind of hack to make Table captions work mostly as
they are documented to, so that {counter:table-number} seems to
act like a substitution.  This hack should be removed when full
attribute value substitution is done.  (This approach has the
benefit of localizing this hack until it can be replaced.)

Of course, this will be incompatible with other uses of the
table-counter counter (because it isn't actually the same counter),
and won't support anything beyond the most trivial substitutions.

While here removed a couple of unused functions.

Fixes #712

### DIFF
--- a/pkg/renderer/sgml/table.go
+++ b/pkg/renderer/sgml/table.go
@@ -14,7 +14,6 @@ func (r *sgmlRenderer) renderTable(ctx *renderer.Context, t types.Table) (string
 	caption := &strings.Builder{}
 
 	number := 0
-	title := r.renderElementTitle(t.Attributes)
 	fit := "stretch"
 	frame := t.Attributes.GetAsStringWithDefault(types.AttrFrame, "all")
 	grid := t.Attributes.GetAsStringWithDefault(types.AttrGrid, "all")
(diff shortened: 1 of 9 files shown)


pkg/renderer/sgml/table.go
```diff
@@ -42,21 +41,20 @@ func (r *sgmlRenderer) renderTable(ctx *renderer.Context, t types.Table) (string
...
-				Title:       title,
-			})
-			if err != nil {
-				return "", errors.Wrap(err, "unable to format table caption")
-			}
+		c, ok := t.Attributes.GetAsString(types.AttrCaption)
+		if !ok {
+			c, _ = ctx.Attributes.GetAsString(types.AttrTableCaption)
 		}
+
```
xcoulon, line 48: should we wait for https://github.com/bytesparadise/libasciidoc/pull/715 to be merged to refactor this part, or address it later? In other words, can this PR only address the `table-caption` attribute?

gdamore: Yes, just merge this -- there are no dependencies.

xcoulon: sorry, I'm not following you... The comment below says that there is a "very primitive and incomplete replacement of the counter attribute only." 
Is it something that will be refactored/improved with #715?

gdamore: Work has to occur after #715 to support attribute values that themselves contain attribute references.  (Meaning recursive attribute substitution.)  That is required -- and will be something done *after* #715 -- 715 is just enabling work to make that happen, and the approach here includes work that will be retained, but also a hack that will be replaced with real attribute substitution later.

gdamore: (Didn't mean to resolve the conversation.)

More precisely we need #604 as well as #715 before we can eliminate this hack.

xcoulon: ok, I see.

xcoulon: thanks for the clarification

PR #723, opened 2020-07-13
bug(renderer): source highlighter should get unadulterated string

### LINKED ISSUE #721: source highlighter should get unadulterated string
The content passed to the source highlighter code winds up having entities injected, which confused the output.  So when I pass this:

```
[source,c]
----
#include <stdio.h>

printf("Hello World!\n");
----
```
I get this:

```
#include &lt;stdio.h&gt;

printf(&#34;Hello World!\n&#34;);
```

gdamore:
Ths fix for this involved some refactoring of the highlighting work
to separate out the calloouts (we embed them using reserved Unicode so
that we can find them again and replace them inline).

Chroma automatically escapes the code for HTML safetly already, so
we defer to it's rules, when we use it.  We still have to do this
ourselves when not using Chroma.

Fixes #721


gdamore:
Any feedback on this one?

xcoulon:
> Any feedback on this one?

I plan to review it later today

xcoulon:
thanks @gdamore!

PR #730, opened 2020-07-17
bug(renderer): Section ID should include the icon class

### LINKED ISSUE #691: Section ID should include the icon class
For example:
```
== icon:warning[] or what icon:note[] to do
```
should render as 
```
<div class="sect1">
<h2 id="_warning_or_what_note_to_do"><span class="icon">[warning]</span> or what <span class="icon">[note]</span> to do</h2>
<div class="sectionbody">

</div>
</div>
```

(note the `id="_warning_or_what_note_to_do"` in the `<h2>` element)

gdamore:
Fixes #691

(This also honors alt if its set on the icon, and uses that instead of the icon class).

### DIFF
--- a/pkg/parser/icon_test.go
+++ b/pkg/parser/icon_test.go
@@ -138,7 +138,7 @@ var _ = Describe("icons", func() {
 					Blocks: []interface{}{
 						types.Section{
 							Attributes: types.Attributes{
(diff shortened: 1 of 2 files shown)


xcoulon:
thanks @gdamore !

Issue #638, opened 2020-06-24
Consider using go generate for parser generation


gdamore:
The current framework for this needs "make" and other tooling which makes it a bit inconvenient to work with from a Windows environment.

Using a go generate line, this would be far easier coming from a a pure Go development environment.

moorereason:
In a similar vein, check out [`mage`](https://github.com/magefile/mage) as a pure Go replacement for `make`.  Hugo's [been using it](https://github.com/gohugoio/hugo/blob/master/magefile.go) for a few years now.

gdamore:
I don't think "make" is ever really needed.  Go generate is more than sufficient for most cases.

gdamore:
I have a trivial change with doc updates for this.  PR coming soon.

xcoulon:
> I don't think "make" is ever really needed. Go generate is more than sufficient for most cases.

it is, and I intend to keep using it ;)

gdamore:
Which are you referring to?  Make or go generate?

xcoulon:
> Which are you referring to? Make or go generate?

`make` is my preferred choice. Take a look at the `make generate` and `make generate-optimized` goals in particular: https://github.com/bytesparadise/libasciidoc/blob/master/make/go.mk#L34-L46 
If a user wants to generate the optimized parser, she'll have to also specify all the alternate entrypoints. In this case, `make generate-optimized` is way simpler than the whole command. And entrypoints are subject to changes as we refactor the code (recently with the substitutions for example), which means that https://github.com/bytesparadise/libasciidoc/blob/master/make/go.mk is a good way to share the changes.

Again, for local testing, `go generate ./...` may be enough, but when submitting a PR, the parser will need to be built with all the required settings.

gdamore:
Please see the updated PR.  The detailed command is part of the go generate file content.  So instead of a make file we have a generate.go.  It includes all the settings.

PR #731, opened 2020-07-17
chore(docs): Update the installation instructions

gdamore:
chore(build): Use go generate

This provides a simpler go generate (meaning developers no longer
need to use "make", which is a qualify of life improvement for
developers working on non-UNIX systems), and updates the instructions
to reflect the trivial installation steps.  (Also details about using
go generate for developers are added to the CONTRIBUTING document.)

This also coincidentally includes updates to the latest dependencies,
which has the benefit of eliminating the protobuf dependency.

Fixes https://github.com/bytesparadise/libasciidoc/issues/638

### DIFF
--- a/CONTRIBUTE.adoc
+++ b/CONTRIBUTE.adoc
@@ -29,6 +29,23 @@ Don’t get discouraged if you don't get an immediate response, this is a side-p
 
 Testing existing features is a good way to contribute to the project, too!
 
+== Modifying The Grammar
+
+If you're working on the code, and want to modify the grammar (parser/parser.peg), you
+will need to install the https://github.com/mna/pigeon[pigeon] grammar generator:
+
+    $ go get -u github.com/mna/pigeon
+
+Then if you're updating the grammar, you then will need to use go generate:
+
+   $ go generate ./...
+
+(This is made possible by the pkg/parser/generate.go file.
+See that file for details on the flags used with pigeon.)
+
+Then you can go build as per normal:
+
+   $ go build ./...
 
 == Benchmarking
 
--- a/pkg/parser/generate.go
+++ b/pkg/parser/generate.go
@@ -1,0 +1,3 @@
+package parser
+
+//go:generate pigeon -optimize-parser -alternate-entrypoints AsciidocRawDocument,RawFile,TextDocument,DocumentRawBlock,FileLocation,IncludedFileLine,InlineLinks,LabeledListItemTerm,NormalBlockContent,NormalParagraphContent,VerseBlockContent,MarkdownQuoteBlockAttribution,InlineElements -o parser.go parser.peg
--- a/README.adoc
+++ b/README.adoc
@@ -77,11 +77,11 @@ Libasciidoc provides 2 functions to convert an Asciidoc content into HTML:
 
 1. Converting an `io.Reader` into an HTML document:
 
-    ConvertToHTML(r io.Reader, output io.Writer, config configuration.Configuration) (types.Metadata, error) 
+    Convert(r io.Reader, output io.Writer, config configuration.Configuration) (types.Metadata, error) 
 
 2. Converting a file (given its name) into an HTML document:
 
-   ConvertFileToHTML(output io.Writer, config configuration.Configuration) (types.Metadata, error)
+   ConvertFile(output io.Writer, config configuration.Configuration) (types.Metadata, error)
 
 where the returned `types.Metadata` object contains the document's title which is not part of the generated HTML `<body>` part, as well as the other attributes of the source document.
 
@@ -104,7 +104,7 @@ var tmpl = template.Must(t.Parse(tmplStr))
 
 output := &strings.Builder{}
 content := strings.NewReader(`example::hello world[suffix=!!!!!]`)
-libasciidoc.ConvertToHTML(context.Background(), content, output, renderer.WithMacroTemplate(tmpl.Name(), tmpl))
+libasciidoc.Convert(content, output, renderer.WithMacroTemplate(tmpl.Name(), tmpl))
 ```
 
 == How to contribute
--- a/README.adoc
+++ b/README.adoc
@@ -53,8 +53,11 @@ Using `-b` (or `--backend`) the following formats are supported:
 
 == Installation
 
+This is a standard Go package, and it installs like you might expect.
+
+To install it in your `$GOPATH` (assuming `$GOPATH/bin` is on your path), just use `go install`:
+
     $ go get -u github.com/bytesparadise/libasciidoc
-    $ make install
 
 == Usage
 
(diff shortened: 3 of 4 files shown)


CONTRIBUTE.adoc
```diff
@@ -29,6 +29,20 @@ Don’t get discouraged if you don't get an immediate response, this is a side-p
 
 Testing existing features is a good way to contribute to the project, too!
 
+== Modifying The Grammar
```
xcoulon, line 32: I understand that it can be a bit of a pain for some users to install `make`, but it would be fair to provide these instructions as an _alternative_ to the `make generate`, `make build` and `make install`, not as the preferred way.

gdamore:
I thought I did have it generate an optimised parser.

Get Outlook<https://aka.ms/qtex0l> for iOS
________________________________

gdamore:
I thought I did this so that it does in fact generate an optimised parser by default.  I see no reason to not do that every time.  (A few dozen milliseconds at build time difference maybe?)

gdamore:
It looks like the go generate file didn't get added (forgot to git add).  Will follow up shortly.

xcoulon:
> It looks like the go generate file didn't get added (forgot to git add). Will follow up shortly.

yes, I was actually wondering a minute ago how the code generation could be triggered without any additional `//go:generate` in your PR ;)

gdamore:
Should be fixed now.

CONTRIBUTE.adoc
gdamore: Please review the updated PR.  I really think make should *not* be the preferred way.

pkg/parser/generate.go
```diff
@@ -0,0 +1,3 @@
+package parser
+
+//go:generate pigeon -optimize-parser -alternate-entrypoints AsciidocRawDocument,RawFile,TextDocument,DocumentRawBlock,FileLocation,IncludedFileLine,InlineLinks,LabeledListItemTerm,NormalBlockContent,NormalParagraphContent,VerseBlockContent,MarkdownQuoteBlockAttribution,InlineElements -o parser.go parser.peg
```
xcoulon, line 3: Ok, I see. That's actually what I was missing in the earlier version of the PR ;) 
In this case, we should update the `generate-optimized` goal in https://github.com/bytesparadise/libasciidoc/blob/master/make/go.mk#L40-L46 to call `go generate ./...` instead, so we don't have 2 places where list of `alternate-entrypoints` need to be maintained.

(Note that for now I'd like to keep the `make generate` goal because I use it in my daily testing workflow: it keeps my code/generate/test workflow pretty fast.)

gdamore: Sure -- but this also helps me ... I always have to shell out to WSL to generate the grammar because I don't have a version of "make" that understands your makefiles on Windows.  (Your makefiles are GNU make, and not stock simple make, which adds yet further onus on folks who have to use them.)

Do you have any use for a separate generate rule, or do you think we could just consolidate those to have and generate and generate-optimized both generate the optimized parser?

xcoulon: As I said:
1. I understand it can be complicated for you on Windows, so I'm fine with using `go generate ./...` under the hood.
2. This PR should refactor the `generate-optimized` goal to use the `go generate ./...` command
3. For now, I'm used to `make generate` because it's faster in my workflow (code/generate/test) so I would like to keep this specific goal as-is. However, I always run `make test` before submitting a PR anyways, so the parser will be optimized at this point.

Does it clarify?

CONTRIBUTE.adoc
```diff
@@ -29,6 +29,20 @@ Don’t get discouraged if you don't get an immediate response, this is a side-p
...
+
+If you're working on the code, and want to modify the grammar (parser/parser.peg), you
+will need to install the https://github.com/mna/pigeon[pigeon] grammar generator:
+
+    $ go get -u github.com/mna/pigeon
+
+Then if you're updating the grammar, you then will need to use go generate:
```
xcoulon, line 39: maybe it's worth adding a link to the `generate.go` file so new users can see how it works?

pkg/parser/generate.go
gdamore: Yes, thanks.  I'm surprised you can tell the difference in speed between generate and generate-optimized.  I haven't measured it, but its far below my threshold of notice.

xcoulon: Hmm, I'll have to check again then, but that was my feeling... 
Anyways, I believe that your main request is being able to use `go generate` and mine is to make sure we don't have to maintain the `alternate-entrypoints` in 2 different places. Once we have the `make generate-optimized` goal using `go generate` under the hood, I will give it another try and maybe remove the `generate` goal afterwards. 

Also, I wonder what the Windows VM on GitHub Actions use, because the `make` commands work fine there 🤔

xcoulon:
@gdamore could you also fix the conflicts with the `go.mod` and `go.sum` files, please?

pkg/parser/generate.go
gdamore: I think they have a full CYGWIN environment.  Which also means GNU make.

Btw, GNU make is *not* portable make.  That syntax isn't understand by NMake (the Visual studios make), pmake (the FreeBSD make utility), or make(1S) as it is found on illumos (though some distributions ship gnumake instead because a lot of people use GNUisms in their makefiles.)

For a pure Go project, its nice to not require any tooling other than what you can get via go get.  It reduces friction.

(For the record, I do use make-like utilities in some of my other projects, but for those I'm using mostly CMake and ninja, which are very nice but completely inappropriate for a Go project.)

xcoulon: Sure, I totally understand your concerns about the friction! 
We use `make` on our projects but we're all on Linux or macOS, so it works in our cases, which is also why I started this project with the tools I'm most familiar with. 
That being said, I'm fine with make your life (and others') easier here, and maybe we can remove all the `make` stuff in the future. Just one thing at a time, ok? ;)

gdamore:
I removed the go mod stuff, because we don't need to update those.  (Common source of conflict.)

I've also added a commit to address #737

pkg/parser/generate.go
xcoulon: > Hmm, I'll have to check again then, but that was my feeling...
> Anyways, I believe that your main request is being able to use `go generate` and mine is to make sure we don't have to maintain the `alternate-entrypoints` in 2 different places. Once we have the `make generate-optimized` goal using `go generate` under the hood, I will give it another try and maybe remove the `generate` goal afterwards.
> 
> Also, I wonder what the Windows VM on GitHub Actions use, because the `make` commands work fine there 🤔

@gdamore last time I checked there were actually barely no difference of execution time between `make generate` and `make generate-optimized`. I'll refactor the `go.mk` later in a subsequent PR

xcoulon:
thanks @gdamore ! 🙌

PR #726, opened 2020-07-14
feature(parser): custom subs on paragraphs

xcoulon:
Support for one or more custom substitutions
Starting with Rawline elements which are parsed
and result in other elements. Subsequent substitutions
only apply on the StringElements.
At the end of the substitutions, the lines are preserved

Also, fix issues with single line comments which should not
start with spaces before the `//` marker

Fixes #597

Signed-off-by: Xavier Coulon <xcoulon@redhat.com>

PR #740, opened 2020-08-01
bug(parser): failure to convert nested elements

### LINKED ISSUE #739: failure to convert nested elements
I have content that looks like:

```
`nng_socket_get()`::
This function is untyped and can be used to retrieve the value of any option.
The caller must store a pointer to a buffer to receive the value in _val_,
and the size of the buffer shall be stored at the location referenced by
_valszp_.
+
When the function returns, the actual size of the data copied (or that
would have been copied if sufficient space were present) is stored at
the location referenced by _valszp_.
If the caller's buffer is not large enough to hold the entire object,
then the copy is truncated.
Therefore the caller should check for truncation by verifying that the
returned size in _valszp_ does not exceed the original buffer size.
+
It is acceptable to pass `NULL` for _val_ if the value in _valszp_ is zero.
This can be used to determine the size of the buffer needed to receive
the object.
+
TIP: It may be easier to use one of the typed forms of this function.
```

When converting this, libasciidoc is tripped up by the new serializeParagraph code, which assumes that these are all RawLines.  Because this contains some embedded stuff, we wind up with the following paragraphs actually being mishandled, as the content isn't a rawline, but rather a []interface{} (which is made up of raw lines I think):

`unexpected type of element while serializing a paragraph: '[]interface {}'`

It actually is tripping up on the thing following the first + sign.

gdamore:
Fixes #739

### DIFF
--- a/pkg/parser/paragraph_test.go
+++ b/pkg/parser/paragraph_test.go
@@ -856,5 +856,91 @@ a foo image:foo.png[]`
 				Expect(ParseDocument(source)).To(MatchDocument(expected))
 			})
 		})
+
+		Context("list paragraphs", func() {
+
+			It("inline image within a quote", func() {
+				source := `item::
+This is the first line of the first paragraph.
+This is the second line of the first paragraph.
++
+This is the first line of the continuation paragraph.
+This is the second line of the continuation paragraph.
++
+This is the next continuation paragraph.
++
+TIP: We can embed admonitions too!
+`
+				expected := types.Document{
+					Elements: []interface{}{
+						types.LabeledList{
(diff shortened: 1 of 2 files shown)


pkg/parser/paragraph_test.go
```diff
@@ -856,5 +856,91 @@ a foo image:foo.png[]`
 				Expect(ParseDocument(source)).To(MatchDocument(expected))
 			})
 		})
+
+		Context("list paragraphs", func() {
+
+			It("inline image within a quote", func() {
```
xcoulon, line 862: I'll rename that test in #726, its title does not reflect the content of `source`

PR #754, opened 2020-09-13
feature(renderer): inline images deserve links, too

pjanx:
They are used in libasciidoc's own README.adoc for badges and I wanted to use them in mine, without working around it with line noise:

```
perl -pe 's|\b(image:.*?\[.*?),\s*link=([^,]*?)(\])|+++<a href=$2>+++\n$1$3\n+++</a>+++|gx'
```

So that I'm not just bitching in issues. :) This was easy enough to add, even if tiring because of two super similar backends. Also notice that the templates are formatted differently.

### DIFF
--- a/pkg/renderer/sgml/html5/image.go
+++ b/pkg/renderer/sgml/html5/image.go
@@ -9,5 +9,5 @@ const (
 {{ else }}
 {{ end }}</div>
 `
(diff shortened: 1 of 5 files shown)


gdamore:
Looks pretty good to me... sorry for missing this in the original work.

xcoulon:
thanks for addressing this issue, @pjanx 🙌

PR #755, opened 2020-09-26
feat(parser): support custom substitutions on delimited blocks

### LINKED ISSUE #558: Custom substitutions on blocks
see how to customize the substitutions on a block. Eg:
```
[source,java,subs="verbatim,quotes"] 
----
System.out.println("Hello *bold* text"). 
----
```

https://asciidoctor.org/docs/user-manual/#applying-substitutions

xcoulon:
applying the same logic as with paragraphs to apply default and
custom substitutions on delimited blocks. This means that
delimited blocks are processed in multiple steps, but each step
has a much simpler grammar.

Also, quite a lot of refactoring and code reorganization for the
delimited block tests in `pkg/parser`.

See #749

Fixes #558


xcoulon:
finally! 😅

PR #819, opened 2021-02-13
refactor(parser): build with '-optimize-grammar' flag

xcoulon:


xcoulon:
`benchstat` report when comparing `master` branch against this PR:
```
name           old time/op    new time/op    delta
Libasciidoc-8     171ms ± 1%     150ms ± 3%  -12.63%  (p=0.000 n=10+9)

name           old alloc/op   new alloc/op   delta
Libasciidoc-8    57.9MB ± 0%    52.8MB ± 0%   -8.74%  (p=0.000 n=9+10)

name           old allocs/op  new allocs/op  delta
Libasciidoc-8      925k ± 0%      904k ± 0%   -2.34%  (p=0.000 n=10+10)
```

Generating `pkg/parser/parger.go` takes much longer, and the file is much bigger, though (18.7k -> 233.9k lines). We'll see how it affects the daily work.

PR #828, opened 2021-02-28
refactor(parser): parse quoted text attribute once

### LINKED ISSUE #827: Avoid parsing quoted text attributes multiple times
Quoted text attributes such as `[.myrole]*cookie*` are parsed for each type of quoted text (bold, italic, etc.). 
Parsing a single time should be enough.

xcoulon:
Do not repeat parsing of attributes for each type of quoted text

Fixes #827

### DIFF
--- a/pkg/parser/document_processing_apply_substitutions.go
+++ b/pkg/parser/document_processing_apply_substitutions.go
@@ -63,8 +63,8 @@ func applySubstitutions(ctx substitutionContext, elements []interface{}) ([]inte
 	}
 	result := make([]interface{}, len(elements))
 	for i, e := range elements {
-		log.Debugf("applying substitution on attributes of element of type '%T'", e)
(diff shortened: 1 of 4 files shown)


xcoulon:
FWIW, the `benchstat` results when running `make bench` on this branch compared to `master`:
```
name           old time/op    new time/op    delta
Libasciidoc-8     135ms ± 1%     129ms ± 6%   -4.58%  (p=0.001 n=10+9)

name           old alloc/op   new alloc/op   delta
Libasciidoc-8    52.4MB ± 0%    44.6MB ± 0%  -14.96%  (p=0.000 n=10+10)

name           old allocs/op  new allocs/op  delta
Libasciidoc-8      903k ± 0%      837k ± 0%   -7.32%  (p=0.000 n=9+10)
```

PR #847, opened 2021-04-09
Update installation instructions

dflock:
Add `make install`

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -52,13 +52,16 @@ Using `-b` (or `--backend`) the following formats are supported:
 
 == Installation
 
-This is a standard Go package, and it installs like you might expect.
+To build libasciidoc and make it available on the command line, do this:
 
-To install it in your `$GOPATH` (assuming `$GOPATH/bin` is on your path), just use `go install`:
+    $ git checkout github.com/bytesparadise/libasciidoc
+    $ cd libasciidoc
+    $ make install
+    $ sudo ln -s  ./bin/libasciidoc /usr/bin/libasciidoc
+
+Libasciidoc is a standard Go package, so if you want to use it as a library in your Go project & install it in your `$GOPATH` (assuming `$GOPATH/bin` is on your path), just use `go install`:
 
     $ go get -u github.com/bytesparadise/libasciidoc
-    $ cd $GOPATH/src/github.com/bytesparadise/libasciidoc
-    $ make install
 
 == Usage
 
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -57,6 +57,8 @@ This is a standard Go package, and it installs like you might expect.
 To install it in your `$GOPATH` (assuming `$GOPATH/bin` is on your path), just use `go install`:
 
     $ go get -u github.com/bytesparadise/libasciidoc
+    $ cd $GOPATH/src/github.com/bytesparadise/libasciidoc
```
xcoulon, line 60: I think we could replace the `go get` command with something like `git checkout github.com/bytesparadise/libasciidoc`. The `go get` is more for users who want to add the library to their project, whereas the `make install` is to have `libasciidoc` available as a command

xcoulon (commented):
thanks for this PR @dflock! I left a comment to suggest a small change

README.adoc
dflock: I've pushed a new commit that updates this - let me know if that's what you meant?

README.adoc
```diff
@@ -52,9 +52,14 @@ Using `-b` (or `--backend`) the following formats are supported:
...
 == Installation
 
-This is a standard Go package, and it installs like you might expect.
+To build libasciidoc and make it available on the command line, do this:
 
-To install it in your `$GOPATH` (assuming `$GOPATH/bin` is on your path), just use `go install`:
+    $ git checkout github.com/bytesparadise/libasciidoc
+    $ cd libasciidoc
+    $ make install
+    $ sudo ln -s  ./bin/libasciidoc /usr/bin/libasciidoc
```
xcoulon, line 60: the `make install` goal already copies the binary into `$GOPATH/bin` so I think that we should ask the users to ensure that `$GOPATH/bin` in `$PATH`, instead of making a symlink into a directory of `$PATH`. Also, `sudo ln -s ./bin/libasciidoc /usr/bin/libasciidoc` won't work for Windows users :(

README.adoc
```diff
@@ -52,9 +52,14 @@ Using `-b` (or `--backend`) the following formats are supported:
...
+
+Libasciidoc is a standard Go package, so if you want to use it as a library in your Go project & install it in your `$GOPATH` (assuming `$GOPATH/bin` is on your path), just use `go install`:
```
xcoulon, line 62: in that paragraph, I would simply refer to the case where the user wants to embed the library in her own project. Ib which case, it's `go get`, not `go install`

xcoulon (changes requested):
Thanks for your PR update and my apologies for the late review @dflock. 
I've left some further comments.

xcoulon:
replaced by https://github.com/bytesparadise/libasciidoc/pull/966

xcoulon:
my apologies for not taking time to address this PR earlier @dflock :/ 
I've added a few changes in #966 and it's merged now

Issue #974, opened 2022-03-10
Add plugin support


rxt1077:
A lot of what makes AsciiDoc so powerful is the variety of plugins available for it. I propose adding a simple plugin framework where golang plugins export functions that are executed at different phases of the `Convert` process.

PR #975, opened 2022-03-10
Plugin support

rxt1077:
Addresses #974.

This is a rudimentary version of plugin support for libasciidoc. It adds a command-line option to the binary and runs a PreRender function from plugins just before the document is rendered. Plugins are golang plugins with the `PreRender` symbol exported.


xcoulon:
thanks for this PR @rxt1077 
I need to take a closer look at Go plugins, since they seem to have quite a few limitations :/ 
Also, could you include some tests to show and verify how it works? This may require some additional goals in the makefile, though 🤔

rxt1077:
Thanks and I understand. Go plugins do seem limited and it's worth exploring other options. This is less of a "best" solution and more of a "possible" solution. I appreciate you taking a look and please don't feel pressured towards any action. I'll see if I can create some tests, I realize that part of my PR is missing.

PR #990, opened 2022-03-20
Fallback symlink command doesn't work

dflock:
If run as-is, the current symlink command will create a broken link:

```
$ sudo ln -s  ./bin/libasciidoc /usr/bin/libasciidoc
$ file /usr/bin/libasciidoc
/usr/bin/libasciidoc: broken symbolic link to ./bin/libasciidoc
```

This command works and produces a working absolute link:

```console
$ sudo ln -s "$PWD/bin/libasciidoc" /usr/bin/libasciidoc
```

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -60,7 +60,7 @@ To build libasciidoc and make it available on the command line, do this:
 
 If `$GOPATH/bin` is already in `$PATH`, then you should be good. Otherwise, for Linux and macOS users, you can run the following command:
  
-    $ sudo ln -s  ./bin/libasciidoc /usr/bin/libasciidoc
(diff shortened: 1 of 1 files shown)


xcoulon:
oh thanks @dflock, i did not notice that error in the doc 🤦‍♂️ 
however, I'm getting this error on my mac:
```
$ ln -s  ./bin/libasciidoc /usr/bin/libasciidoc
ln: /usr/bin/libasciidoc: Operation not permitted
```
but it works with `sudo ln -s  $PWD/bin/libasciidoc /usr/local/bin/libasciidoc` instead. 
Depending on your OS and if the command above also works for you, could you update the PR accordingly?

xcoulon:
thanks for this PR @dflock! 🙌

PR #994, opened 2022-04-18
Issue992 benchstat comments

xcoulon:
- chore(bench): display bench stats as comment in PR
- fix syntax
- use var for diff output
- phony makefile goal
- git checkout
- ref name
- symbolic ref or current commit
- hide git command and show path to diff report
- bench with Go 1.17.x only
- echo instead of cat
- remove cat in GH action workflow
- echo/cat
- edit OUTPUT
- cleanup
- do not use explicit token
- change event type
- split build/test vs lint workflows, increase linter timeout
- rename workflow and jobs
- revert workflow changed
- fix error
- grant permissions to write on PR


xcoulon:
closed in favor of https://github.com/bytesparadise/libasciidoc/pull/995

PR #997, opened 2022-04-22
fix git clone command in README

stevegt:


xcoulon (approved):
thanks for catching this error @stevegt 🙌

PR #1100, opened 2022-11-08
Don't lowercase section titles in parser

### LINKED ISSUE #1063: Natural internal cross reference gets changed to lowercase outside of the template
I am building a custom backend based of the `sgml` group. Basically a custom copy of the xhtml one. I was able to build a document setting:

```
:idprefix:
:idseparator: -

See <<My natural Link>>

== My natural Link
```

With those attributes set, I was expecting my cross reference to be something like `#My-natural-Link` but it got automatically lower cased. 
The casing information is lost by the time you get to the template rendering. From the example above, you can't ensure that upcasing the first letter is enough.

Is there a way to not automatically lowercase the reference in the code and leave it to the template? You probably would have to add a `ToLower` funcMap for the template to be able to do it.

Asciidoctor doesn't have an extra attribute to control casing of the section IDs so it is probably not worth introducing one if you leave to the template but that would do it too.
https://docs.asciidoctor.org/asciidoc/latest/sections/id-prefix-and-separator/

The code that lowercases the ID is here:
https://github.com/bytesparadise/libasciidoc/blob/e12e66011b03f24f05e04013767a867127be2bdb/pkg/types/non_alphanumerics_replacement.go#L68

Removing that strings.ToLower fixes things for me but breaks a few tests obviously. Let me know how you would like to proceed. I could draft a PR based on that.

DavidGamba:
Defer lowercase to templates instead

Fixes #1063


xcoulon:
Thanks for this PR, @DavidGamba!

xcoulon (approved):

